PR #46, opened 2016-11-29
Fixes some typos in breakIdeas.js

alexalekseyenko:
Tried to change as little as possible and preserve the original meaning.

Thanks for the awesome app -- I'm using it on the daily!


hovancik:
Thanks a lot for the contribution ;] If you have some new break ideas that you would like to add, feel free to make new PR and we can discuss them.

PR #64, opened 2017-03-03
Quick-and-dirty dark mode

alexalekseyenko:
I've been moving increasingly toward a dark-themed environment and Stretchly is one of the few remaining applications I use all the time which doesn't have a dark mode. This is a really crude pass at one (mostly to run it by you and see what you think).

On account of the limited space, I had to choose a theme to sacrifice and "splash of cappuccino" got the axe. Mostly an arbitrary choice, but I figured we already have one coffee-themed color scheme, and this one provides the least contrast for the light text.

The background color is copied from this excellent vim color scheme:
https://github.com/w0ng/vim-hybrid


hovancik:
I like it ;] Nice color. I've made some changes so the app looks better. 
Do you see some settings here to let me push to your branch? Not sure how can I push to your branch. 
I would like you to try my changes.

alexalekseyenko:
Looks great 👍  I like the changes.

PR #65, opened 2017-03-03
Add monochrome tray icon

### LINKED ISSUE #60: Support for Monochrome icon on OS X
Could you add support for monochrome icons on OS X so the icon doesn't look ugly in the menu bar. Also a HighDPI (2x) icon would be appreciated, as it looks quite pixelated on Retina displays.

alexalekseyenko:
fixes #60


app/main.js
```diff
@@ -60,11 +60,15 @@ function displaysY (height = 600) {
 }
 
 function createTrayIcon () {
```
hovancik: I think only macOS should be monochrome. On Linux and Windows we should have full-color stretchly logo

alexalekseyenko: Haven't had a chance to test it on Linux, but that should restore the original behavior.

hovancik: so you think that Windows should be also monochrome?

alexalekseyenko: Doh! I forgot that I generated the .ico file from the monochrome icon. Thank you for catching that

hovancik: Thanks, tested on Linux as well and looks good!

PR #73, opened 2017-03-23
Shwe

shweaye810:
make settings more clear with numbers and operators.

### DIFF
--- a/app/break.html
+++ b/app/break.html
@@ -7,6 +7,7 @@
   </head>
   <body>
     <div class="microbreak">
+      <h2> Time to take a break! </h2>
       <div class="break-idea"></div>
       <div class="break-text"></div>
       <a id="close" title="Ctrl/Cmd + x">...nah, I'll just keep going on</a>
--- a/app/settings.html
+++ b/app/settings.html
@@ -8,13 +8,9 @@
     <div class="settings">
       <h2>microbreaks</h2>
       <p>
-        <span id="microbreakDurationMinus" class="pointer">-</span>
-        <span id="microbreakDuration"></span>
-        <span id="microbreakDurationPlus" class="pointer">+</span>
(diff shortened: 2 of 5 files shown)


app/settings.html
```diff
@@ -23,13 +19,9 @@
       </p>
       <h2>breaks</h2>
       <p>
-        <span id="breakDurationMinus" class="pointer">-</span>
-        <span id="breakDuration"></span>
-        <span id="breakDurationPlus" class="pointer">+</span>
+        <input type="number" id="breakDuration" min="0" max="59" step="1" value=5>
```
hovancik: Did you check how it looks? It's different after your changes. 

screen shot 2017-03-23 at 21 13 07

Also step should not be 1. Check how previous version was. 

That is why I am doing it how I am doing it. It looks better and I do not need to write css for inputs.

app/break.html
```diff
@@ -7,6 +7,7 @@
   </head>
   <body>
     <div class="microbreak">
+      <h2> Time to take a break! </h2>
```
hovancik: I think we do not need this, we can separate between break and microbreak very easy: 
- microbreak - one short sentence
- break - title and sentence

I prefer not to introduce this.

shweaye810: Oh, ok. I didn't know that. I think it should mention somewhere.

app/settings.html
shweaye810: Steps can be fixed very easily. While I was using the app, I was confused by the positive and negative sign, I wasn't even sure if I can click it.

shweaye810: Actually, I want to work on showing the timer, but I am just trying out some other code and decided to change it a little.

shweaye810: IMO, the text should be different with something clickable, or else it's confusing.

hovancik:
Project has moved on a bit since this PR: there is time left in break, so I am closing this.

Issue #63, opened 2017-03-02
Time remaining in icon


pmontepagano:
It would be nice if there was an option to make the icon represent how much time there is left until next break. I was thinking in a circle that fills/empties, similar to MacOS progress icon. 
![240_degrees_filled1600](https://cloud.githubusercontent.com/assets/1820549/23510086/f0c1677c-ff36-11e6-9cd5-2e0db16b3677.png)

hovancik:
Thanks, I like the idea. Will check if that's possible with electron

hovancik:
So this is possible, but I will need help with creating images  (of tray icon) in different stage of the animation for this. This might take some time.

sxmanton:
I don't think you need to create a ton of images. It seems feasible to me that you could use a similar approach as in this project: https://github.com/asheb/timer
It appears to create a hidden window where it can use a canvas to programatically draw the arcs, which is then retrieved via nativeImage.createFromDataURL

I think you could just set the current icon in the background of the canvas, then draw the arcs on top with some transparency to achieve the desired effect.

Personally though, while nice and pretty, I don't think the circular timer visualization would be as useful to me as just the minutes remaining being displayed in the tooltip when hovering over the icon.

hovancik:
I think we can close this with https://github.com/hovancik/stretchly/pull/92 ?

sxmanton:
I'd still like to give the animated icon a shot at some point, but I do think the tooltip is sufficient

hovancik:
I am closing it. Thinking about it now, after some time, I think it would probably be obtrusive to always have animated icon in tray. I would not like it, personally. I like your solution the best.

PR #84, opened 2017-04-22
Adds info to tooltip on time till next break or end of pause

sxmanton:
Not quite as stylish as what's suggested in https://github.com/hovancik/stretchly/issues/63 but still useful I think. The choices for time formatting may seem a little strange, but I figure the point isn't to be constantly monitoring the exact time left till a break, but rather just to give a rough idea.

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -436,7 +436,7 @@ function updateToolTip () {
     let breakType = nextBreakType()
     if (breakType) {
       toolTipString += `\n${Utils.formatTillBreak(breakPlanner.scheduler.timeLeft)} to ${breakType}`
-      if (breakType == 'microbreak') {
+      if (breakType === 'microbreak') {
         let breakInterval = breakPlanner.settings.get('breakInterval')
         let breakNumber = breakPlanner.breakNumber % breakInterval
         toolTipString += `\nNext break following ${breakInterval - breakNumber} more microbreak(s)`
--- a/app/main.js
+++ b/app/main.js
@@ -437,7 +437,7 @@ function updateToolTip () {
     if (breakType) {
       toolTipString += `\n${Utils.formatTillBreak(breakPlanner.scheduler.timeLeft)} to ${breakType}`
       if (breakType === 'microbreak') {
-        let breakInterval = breakPlanner.settings.get('breakInterval')
+        let breakInterval = breakPlanner.settings.get('breakInterval') + 1
         let breakNumber = breakPlanner.breakNumber % breakInterval
         toolTipString += `\nNext break following ${breakInterval - breakNumber} more microbreak(s)`
       }
--- a/app/main.js
+++ b/app/main.js
@@ -5,6 +5,8 @@ const AppSettings = require('./utils/settings')
 const defaultSettings = require('./utils/defaultSettings')
 const IdeasLoader = require('./utils/ideasLoader')
 const BreaksPlanner = require('./breaksPlanner')
+const Utils = require('./utils/utils')
+const Scheduler = require('./utils/scheduler')
 
 let microbreakIdeas
 let breakIdeas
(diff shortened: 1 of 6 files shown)


app/main.js
```diff
@@ -67,9 +71,9 @@ function createTrayIcon () {
   } else {
     appIcon = new Tray(iconFolder + '/stretchly_18x18.png')
   }
-  appIcon.setToolTip('stretchly - break time reminder app')
   isPaused = false
   appIcon.setContextMenu(getTrayMenu())
+  toolTipUpdater = setInterval(updateToolTip, 1000)
```
sxmanton: Contemplating whether, since the time shown is already a rough estimate, the toolTipUpdater should also run at a longer interval, maybe 5 or 10 seconds instead of just one.

hovancik:
Hi, thanks for PR. What is the desired behaviour?  

I am trying it and I am not getting any tooltip, except when there is a break or microbreak going on. 

On first microbreak I am getting: 

```
stretchly - break time reminder app
-15s to microbreak
Next break following 1 more microbreak(s)
```

On second microbreak I am getting: 

```
stretchly - break time reminder app
-15s to microbreak
Next break following 2 more microbreak(s)
```
I guess that is not correct? 



So I went and logged tooltip in console. 
When app starts it is: 

```
stretchly - break time reminder app
10m to microbreak
Next break following 1 more microbreak(s)
``` 

That should be 2, right? 

After first break: 

```
stretchly - break time reminder app
10m to microbreak
Next break following 2 more microbreak(s)
```

That should be 0 (or some another text)? 

After 2nd microbreak I am getting: 

```
stretchly - break time reminder app
3m to break
```
That looks good. 

I am going through code, I might make some comments.

app/main.js
```diff
@@ -5,6 +5,8 @@ const AppSettings = require('./utils/settings')
 const defaultSettings = require('./utils/defaultSettings')
 const IdeasLoader = require('./utils/ideasLoader')
 const BreaksPlanner = require('./breaksPlanner')
+const Utils = require('./utils/utils')
```
hovancik: I think we should get this out of `main.js` to make it not so complicated.

app/main.js
```diff
@@ -234,15 +238,15 @@ function pauseBreaks (seconds) {
     finishBreak()
   }
   breakPlanner.pause()
-  if (seconds !== 1) {
-    resumeBreaksTimer = setTimeout(resumeBreaks, seconds)
+  if (milliseconds !== 1) {
+    resumeBreaksScheduler = new Scheduler(resumeBreaks, milliseconds)
```
hovancik: I would try to get `milliseconds` to `breakPlanner.pause()` as parameter and get rid of need to have `Scheduler` in `main.js` 

What do you  think about it?

app/main.js
```diff
@@ -419,6 +423,40 @@ function getTrayMenu () {
   return Menu.buildFromTemplate(trayMenu)
 }
 
+function updateToolTip () {
```
hovancik: Also here we should do something like `breakPlanner.status()`that would return string for tooltip, not doing it in `main.js`

app/main.js
hovancik: Yeah, doesn't need to be 1 second.

sxmanton:
Hi, thanks for your comments. Realized I didn't fully think my way through the break interval logic, a  + 1 fixes it.  I'll work on refactoring it out of main.js later.

app/main.js
sxmanton: I'm thinking maybe change the logic to be update every 10 secs, round to the nearest 5 seconds when under a minute.

hovancik: Looks good to me.

sxmanton:
Closing, see #92

PR #92, opened 2017-05-06
Adds time info to tooltip and fixes a couple bugs

sxmanton:
This is a continuation of #84, but rebased off the most recent commits so as to avoid merge conflicts. As requested, I refactored the majority of the relevant logic into BreaksPlanner. As I was doing so I also noticed a couple small bugs that I squashed. Not sure if you'd prefer those bugfixes be in a separate PR, I can do that if you like, but I felt they were relevant enough to this feature to be included. See changelog and https://github.com/sxmanton/stretchly/commit/ddc4940dce4cc484dde43dae27ee1478e77128d9, https://github.com/sxmanton/stretchly/commit/034265f9d564c717f184160f7317560aba8e5810 for details

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -18,7 +18,6 @@ let settingsWin = null
 let finishMicrobreakTimer
 let finishBreakTimer
 let settings
-let toolTipUpdater
 let toolTipHeader = 'stretchly - break time reminder app'
 
 global.shared = {
@@ -68,7 +67,7 @@ function createTrayIcon () {
     appIcon = new Tray(iconFolder + '/stretchly_18x18.png')
   }
   appIcon.setContextMenu(getTrayMenu())
-  toolTipUpdater = setInterval(updateToolTip, 10000)
+  setInterval(updateToolTip, 10000)
 }
 
 function startProcessWin () {
--- a/app/main.js
+++ b/app/main.js
@@ -124,6 +124,7 @@ function startMicrobreak () {
     microbreakWin.show()
     finishMicrobreakTimer = setTimeout(finishMicrobreak, settings.get('microbreakDuration'))
   })
+  updateToolTip()
 }
 
 function startBreak () {
@@ -159,6 +160,7 @@ function startBreak () {
     breakWin.show()
     finishBreakTimer = setTimeout(finishBreak, settings.get('breakDuration'))
   })
+  updateToolTip()
 }
 
 function finishMicrobreak (shouldPlaySound = true) {
@@ -176,6 +178,7 @@ function finishMicrobreak (shouldPlaySound = true) {
     microbreakWin = null
     breakPlanner.nextBreak.plan()
   }
+  updateToolTip()
 }
 
 function finishBreak (shouldPlaySound = true) {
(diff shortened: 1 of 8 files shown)


hovancik:
Great, I'll have a look at the code ;] Just tested on Debian Linux and seems to work fine. What platform are you on?

sxmanton:
I'm on Windows 10

app/main.js
```diff
@@ -17,9 +17,9 @@ let aboutWin = null
 let settingsWin = null
 let finishMicrobreakTimer
 let finishBreakTimer
-let resumeBreaksTimer
 let settings
-let isPaused = false
+let toolTipUpdater
```
sxmanton: Linter doesn't like this. The reference is unused right now but I was just being wary of setting an interval timer without having a reference to clear it with later. Admittedly I can't think of what scenario the tooltip updater interval would need to be cancelled though

app/main.js
```diff
@@ -320,13 +319,14 @@ function getTrayMenu () {
     type: 'separator'
   })
 
-  if (!isPaused) {
+  if (breakPlanner == null || !breakPlanner.isPaused) {
```
sxmanton: I'm not sure if the null check is necessary here. I checked and it seems as long as the line with `app.on('ready', createTrayIcon)` is after `app.on('ready',loadSettings)` then breakPlanner should already exist.

hovancik: Yeah, `app.on('ready'` is what happens when app starts. Maybe also create function for tool tip updater? Might be more code, but easier to understand what happens when app starts. What do you think? 

I think we can go without checking for null.

app/main.js
```diff
@@ -67,9 +67,8 @@ function createTrayIcon () {
   } else {
     appIcon = new Tray(iconFolder + '/stretchly_18x18.png')
   }
-  appIcon.setToolTip('stretchly - break time reminder app')
-  isPaused = false
   appIcon.setContextMenu(getTrayMenu())
+  toolTipUpdater = setInterval(updateToolTip, 10000)
```
sxmanton: > Maybe also create function for tool tip updater? Might be more code, but easier to understand what happens when app starts. What do you think?

As in move this line into it's own function, as an additional `app.on('ready')` call?

hovancik: Yes, but looking more into it, now I think it's bad idea ;]

hovancik:
So I tested on macOS and tooltip is not being shown. Might be related to https://github.com/electron/electron/issues/3599

sxmanton:
hmm, that's unfortunate. I don't have any good way of testing on a mac... I guess I could install a VM.

hovancik:
I'll check tomorrow. Probably just enable dock icon.

app/main.js
hovancik: So let's remove it?  We can always add it later, if needed.

sxmanton: Done

hovancik:
can you give me pull rights? I'd like to remove stuff for macOS to make it work. or you can do it by removing `app.dock.hide()`

sxmanton:
I removed it, gave you invite for push access for the future though

hovancik:
So I think we can merge this? All clear for me.

sxmanton:
Yeah, if you've tested it on mac and everything seems like it's working right, then I'd say it's good to go.

Issue #59, opened 2017-02-10
Monitor system idle time for natural breaks


mdekstrand:
It'd be fantastic if stretchly could monitor user idle time for 'natural' breaks - if I've been away for the computer for 10 minutes, for example, it should count as my last break and restart the timer when I come back. Ditto for 1-minute rests resetting the microbreak timer.

Both Workrave and Time Out (for Mac) have this feature.

I do not know if Electron exposes the necessary hooks to implement it. It may require platform-specific binary extensions, unfortunately.

hovancik:
I've been looking into that some time ago, there was some node package. Not sure how platform specific.

chuck13:
I found this module (https://github.com/paulcbetts/node-system-idle-time) while trying to add idle detect support, It seemed enough for me.

p.s. another way to add idle detection on OS X http://stackoverflow.com/questions/2922846/detecting-user-activity-on-mac-os-x

hovancik:
that package is for linux only, right? @chuck13

chuck13:
@hovancik no, it's for Linux, OS X and Windows 😃

hovancik:
Ah, I see. I misread the readme ;]

birkany:
But what if I was reading a technical article and looking to an equation for more than 10 minutes? It is not an idle time actually. It may be hard to distinguish this natural breaks. So I may suggest adding a reading mode also, so the application won't try to detect natural breaks and continue counting down when the reading mode is selected. Some users may like to use in reading mode always, if their work is reading intensive.

mdekstrand:
@birkany having it configurable would make sense, particularly since different people have different needs.

I use it almost entirely to manage RSI, with vision breaks as a side benefit, so as long as I am not operating the keyboard or mouse, I am 'taking a break'.

hovancik:
Hey guys, 

just wanted to let you know, that this feature is coming very slow (in fact, I am not working at it now). In those mentioned npm packages, there is stuff I do not understand, so I am a bit like "not wanna use them" right now. 

Anyhow, I am thinking about adding it for macOS as test. How about showing notification, when idle time is detected?

chuck13:
That would be very cool, I'm looking forward to the new feature.

mdekstrand:
I have started working on this in a branch on my fork, but am currently running in to a problem where the Electron process cannot load the system-idle-time bindings.

hovancik:
@mdekstrand how about going with macOS only right now? 

[...]

PR #93, opened 2017-05-06
Idle time support

mdekstrand:
This adds support for 'natural breaks' using [system-idle-time](https://github.com/paulcbetts/node-system-idle-time), working on #59.

I have only tested this code on Windows.

It adds one settting to the settings screen, but I have not updated the screenshot because Windows will be inconsistent with Mac.

### DIFF
--- a/app/breaksPlanner.js
+++ b/app/breaksPlanner.js
@@ -10,6 +10,12 @@ class BreaksPlanner extends EventEmitter {
     this.scheduler = null
     this.isPaused = false
 
+    this.nextBreakDuration = 0
+    this.idleCheckTimer = null
+    this.scheduleStartIdleTime = 0
+    this.inNaturalBreak = false
+    this.lastIdleTime = 0
+
     this.on('microbreakStarted', (shouldPlaySound) => {
       let interval = this.settings.get('microbreakDuration')
       this.scheduler = new Scheduler(() => this.emit('finishMicrobreak', shouldPlaySound), interval, 'finishMicrobreak')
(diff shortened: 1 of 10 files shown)


sxmanton:
Hi, I'm trying to build this and give it a test, however, when I run npm install, I'm getting an error: "paulcbetts/system-idle-time is not in the npm registry". Any ideas what the issue might be?

sxmanton:
Nevermind, I just figured out that somehow I had two different versions of npm installed and one was much lower version. Uninstalled the lower version and it got past that step, although it still failed with "@paulcbetts/system-idle-time@1.04 install: 'node-gyp rebuild' Exit status 1". This issue seems to be referenced here: https://github.com/RocketChat/Rocket.Chat.Electron/issues/335 so I'm now trying some of those suggested steps. Did you have to go through such a process at all? Might be worth adding something to the readme about

app/breaksPlanner.js
```diff
@@ -15,17 +20,31 @@ class BreaksPlanner {
     let interval = this.settings.get('microbreakInterval')
     if (!shouldBreak && shouldMicrobreak) {
       this.scheduler = new Scheduler(this.microbreakFunc, interval)
+      this.nextBreakDuration = this.settings.get('microbreakDuration')
     } else if (shouldBreak && !shouldMicrobreak) {
       this.scheduler = new Scheduler(this.breakFunc, interval * (this.settings.get('breakInterval') + 1))
+      this.nextBreakDuration = settings.get('breakDuration')
```
sxmanton: Should be this.settings instead of settings

app/breaksPlanner.js
```diff
@@ -15,17 +20,31 @@ class BreaksPlanner {
...
     } else if (shouldBreak && shouldMicrobreak) {
       this.breakNumber = this.breakNumber + 1
       let breakInterval = this.settings.get('breakInterval') + 1
       if (this.breakNumber % breakInterval === 0) {
         this.scheduler = new Scheduler(this.breakFunc, interval)
+        this.nextBreakDuration = settings.get('breakDuration')
```
sxmanton: Should be this.settings instead of settings

mdekstrand:
I didn't have any build failures, but did need to use `electron-rebuild` to avoid a load failure.

On Linux it will require the XScreenSaver extension dev files (Ubuntu package `libxss-dev`).

hovancik:
I've updated this with master. No idea if it works, thought.

hovancik:
@mdekstrand could you please explain how your mechanism works? I have hard times understanding it.

hovancik:
seems that author is no longer interested in this so I am closing it. I'm working on new solution at #140

mdekstrand:
fine with me, i haven't had time to look at it any more since submitting :(

PR #97, opened 2017-05-08
Enhance test

### LINKED ISSUE #95: Running tests causes test-missing.json to be deleted
At any rate, any time I run the full test suite, the test-missing.json file is deleted and I have to be careful not to add the deletion into the commit, and git reset it to restore it. Is this an intentional behavior?

The cause seems to be, in the test/settings.js:
```
afterEach(() => {
    if (settings && settings.destroy) {
      settings.destroy()
      settings = null
    }
  })

```
and in app/utils/settings.js:

```
destroy () {
    this.data = null
    fs.unlinkSync(this.settingsFile)
  }
```

What is the reason for this destroy function? I can't find it being called anywhere besides in the test.

hovancik:
we create copy to not worry about git changes to `test-missing.json`

fixes #95

### DIFF
--- a/test/settings.js
+++ b/test/settings.js
@@ -1,11 +1,13 @@
 const chai = require('chai')
 const path = require('path')
+const fs = require('fs')
 // Actual Test Imports
 const Settings = require('./../app/utils/settings')
 const defaultSettings = require('./../app/utils/defaultSettings')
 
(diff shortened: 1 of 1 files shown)


sxmanton:
Thanks :)

PR #116, opened 2017-08-01
Covered version checker.

gromchen:
I wanted to cover `about.js` and `process.js` but it's not possible right now because a new instance of `VersionChecker` is created. Do you consider the use of any framework and DI?


hovancik:
Hi and thanks for the contribution! 

What do you mean by framework and DI?  

Not sure what is wrong with tests, but I would like this PR to be separated in 2, this way it might be easier to figure out what is wrong. I'll add some comments in code.

.eslintrc.json
```diff
@@ -0,0 +1,8 @@
+{
```
hovancik: Adding eslint could be the first PR.

package.json
```diff
@@ -13,7 +13,8 @@
     "dist-all": "build -mwl --x64 --ia32",
     "pack": "build --dir",
     "dist": "build --x64 --ia32",
-    "test": "mocha test"
+    "test": "eslint . && mocha test",
```
hovancik: This should probably be `./node_modules/.bin/eslint` for people who do not have it globally

app/utils/versionChecker.js
```diff
@@ -1,18 +1,15 @@
 class VersionChecker {
-  latest (callback) {
```
hovancik: Could you please explain rationale behind this change? so I understand it better.

test/versionChecker.js
```diff
@@ -0,0 +1,39 @@
+const expect = require('chai').expect
```
hovancik: This would be the 2nd PR. Updating versionChecker and adding test for it.

package.json
gromchen: I don't have eslint globally. Here is an article I found that proves that npm adds node_modules/.bin/ to the PATH environment variable. And tests passed on the build agent.

hovancik: Hm, it doesn't work for me with `eslint .`, only with `./node_modules/.bin/eslint .`

app/utils/versionChecker.js
gromchen: Here was a mix of callback approach and promise. But it's not a big issue, the main problem was that method handled the exception on its own. So users of the method couldn't catch the actual exception and handle it properly. Moreover, they couldn't continue a chain of promises. Also, we could use async await here but it involves an update of the node.

package.json
gromchen: Let me try to test it with different versions of node and npm.

app/utils/versionChecker.js
hovancik: Ah, thanks, I am not that familiar with that stuff so I will try to learn from it. 

About **node**: I am copying what [electron](https://electron.atom.io/) version is using, so right now it's 7.4.0. Next version electron v1.7.0 beta will have 7.9.0.

package.json
gromchen: I switched to Windows and tested it with node v6.9.4 and npm 3.10.6 and it worked. Strange but let's remove it for now from test script since I guess istanbul can't process eslint result. It's still useful to have eslint locally though, then an editor can use it for error checking.

gromchen: Do you use mocha globally? Because I have it locally as well and it works for me.

hovancik: I think it's not the eslint issue, that's why I wanted to create 2 PRs from this. 

You can reference eg. to https://github.com/hovancik/stretchly/blob/master/.travis.yml to see that I am not using `npm test` but `istanbul cover ./node_modules/mocha/bin/_mocha`.

gromchen: I removed eslint for now because I've noticed that if one runs it with old node version it goes crazy. And I suggest adding istanbul to test script.

app/utils/versionChecker.js
gromchen: There was a problem with sinon, it wasn't restored after tests. Usage of sandbox fixed the problem.

gromchen:
About framework/DI I mean that for example currently about.js does not export anything to test, so there are no hooks or seams to insert mocks/stubs/spies. Frameworks are not essential here but we should have something to import in out tests at least.

hovancik:
Ah, I see. Not sure what is the electron way to do this. There is [Spectron](https://github.com/electron/spectron) to do some headless tests. 

I was never really thinking about this, as eg. `about.js` is for html.

Anyway, would you like to add yourself to the README? I can merge this after.

PR #134, opened 2017-09-06
Update README.md

cloo:
Clarified how to find JSON file.


cloo:
Yes, it's  already mentioned. 
But it when I intended to change some settings, I didn't even consider reading about debugging.
So imho, it makes some sense.

Issue #136, opened 2017-09-08
Translate


lvloha:
When do you plan to translate the application? at least you can start with the language of the massages that come out in the breaks.
Or at least put them in an easy file to open and translate.

Very good application, thank you.

hovancik:
Hi and thanks! 
You can edit break ideas manually https://github.com/hovancik/stretchly#editing-breakmicrobreak-ideas

Translations are in my TODO but no idea when I will do it.

Issue #108, opened 2017-07-23
Translations 


jony0008:
Would be great to have stretchly in different languages. I can recommend Crowdin or Weblate for managing the translations.

hovancik:
Hi, this is on my todo list, I'll add your recommendations

aGoogleFan:
I would be glad to help you add Persian/Farsi language, but I donnu how.

hovancik:
@aGoogleFan the easiest would be to create a PR. Do you know how to do it?

aGoogleFan:
unfortunately, I don't know.
i would be appreciate if you link me to an instruction about doing such a job.

hovancik:
@aGoogleFan  what platform are you on? 

github has docs here: https://help.github.com/articles/about-pull-requests/ 

if you are not familiar with git in general, this might be too hard

PR #143, opened 2017-10-04
add a language

wilicw:
I add chinese,and add language selection function, but in the idea that there is a mistake, trouble you to repair, I hope you can use the request

### DIFF
--- a/app/about.html
+++ b/app/about.html
@@ -2,19 +2,16 @@
 <html>
   <head>
     <link rel="stylesheet" type="text/css" href="css/app.css">
+    <meta charset="UTF-8"/>
   </head>
   <body>
     <div class="about">
       <div class="about-name">
         stretchly
       </div>
-      <div>
-        break time reminder app
-      </div>
+      <div id="adoutsecondtitle"></div>
       <div class="about-stretchly">
-        <p>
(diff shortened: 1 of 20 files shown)


hovancik:
Hi and thanks for the input! 

Translations are on my TODO list and some people already opened issues about it: #136, #108. Your implemntation is not bad, but I think we should go with something "nodejs". I'm not familiar myself with nodejs internationalization, but there seems to be some packages: https://github.com/mashpie/i18n-node, https://github.com/i18next/i18next, https://github.com/martinheidegger/i18n-core. Some people recommended https://www.gnu.org/software/gettext/. 

We also need to think about how we wanna translate, as there are also online tools like https://crowdin.com/ or https://weblate.org/en/.

wilicw:
Thank you for your reply, I think I can help you translate into Chinese in the future😀

hovancik:
If you are interested, we could work on this together.

midpoint:
[chinese.zip](https://github.com/hovancik/stretchly/files/1390213/chinese.zip)
This is my production of Chinese package,unpack  it to resources folder.

hovancik:
The interface are now in master, feel free to make a PR with translation file.

PR #180, opened 2018-01-07
Add Bulgarian translation

unholyHub:


hovancik:
Hi and many thanks for the translations! 

Could you please add the Bulgarian language to the settings page? 
It's file called `settings3.html`: 

```
          <option value="bg">Bulgarian in bulgarian language</option>
          <option value="cs">Čeština</option>
          <option value="de">Deutsch</option>
```

PR #182, opened 2018-01-08
Update settings3.html

unholyHub:


hovancik:
Thanks, I've added it to the single PR. #180

PR #188, opened 2018-01-16
Added multi-monitor support for breaks.

### LINKED ISSUE #117: Multiple monitor suggestion
Running on Gnome 3.24

Very nice app, but I wondered if there would be a way to have the window appear on all screens, not just one? Would make it harder to miss if I have mouse focus on one screen but am looking at another.

fdalvi:
Using the Electron Screen API, we launch a BrowserWindow per monitor
if the users preferences are set to do so. Added additional entry
in Settings page 3 for this functionality.

fixes #117


fdalvi:
I'm still unsure about one thing in this implementation - each BrowserWindow sends its own `finish-microbreak` event. Currently, the main process closes all break windows upon receiving this message, and `finish-microbreak` events from other windows are just ignored. However there may be some race conditions in this - so maybe the implementation should close _only_ the window who sent the `finish-microbreak` event (and the last `finish-microbreak` takes care of cleanup)

hovancik:
Thanks for the PR ;] 

Off the Top of My Head: 
- I think this behavior should be turned on by default
- I would image the break idea is shown only on active window ( "active" I mean as was in previous version) and other windows are empty (no text, no timer), or having the same break idea on every window 
- Also I think all windows should close the same time. What race condition do you mean? I think the best would be to have one "active" window as I mentioned before and let others be "dummies". 

What do you think? I hope more people will weight in.

fdalvi:
Hi,

I was off for vacation the past week, so didn't get a chance to get back to this. 

- I agree, I will add it to the default settings
- It should be straightforward to add in a dummy window for the non-primary (or non-active) windows - although I feel like it looks too empty and would prefer the latter (same idea on all displays). Sometimes when the break starts, my cursor is actually on a secondary small display, and having a blank window on my larger monitor just feels weird.
- I actually misunderstood the code when I wrote about the race condition - this is not an issue at all actually!

Currently the easiest thing to do is use the same idea and have identical break screens on all displays - with the timer/buttons mirrored exactly. If you feel like this is too redundant, we can switch to only ideas on all displays, and timer/buttons on active display.

Let me know what you think!

hovancik:
I think you can go ahead with the same idea on every window. My concern is to duplicate timers, as it would add complexity. One timer, as we have it now, should be enough. Or I'm wrong?

fdalvi:
Only one idea is shown on all screens now! 

Also fixed a bug where I accidentally launched one scheduler per screen, but cancelled only the last scheduler when the user ended a break manually. The other schedulers would just complete as if a break was actually in progress. Now I launch only one scheduler regardless of the number of displays!

Regarding the timers, there is only one real timer (in the scheduler) - as far as I understood the timer progress bars in the breaks are just independent animations right?

app/main.js
```diff
@@ -89,16 +89,41 @@ function startPowerMonitoring () {
   })
 }
 
-function displaysX (width = 800) {
+function numDisplays () {
```
hovancik: maybe rename this to something more meaningful? Like `displaysSize`, `numberOfDispalys`?

app/main.js
```diff
@@ -246,13 +297,15 @@ function finishMicrobreak (shouldPlaySound = true) {
...
   }
-  if (microbreakWin) {
+  if (microbreakWins) {
     if (process.platform === 'darwin') {
       // get focus on the last app
       Menu.sendActionToFirstResponder('hide:')
     }
-    microbreakWin.close()
-    microbreakWin = null
+    for (let i = microbreakWins.length - 1; i >= 0; i--) {
```
hovancik: could we create a function out of this?  it would accept array of windows and close them one by one.

hovancik:
note to self: check if cursor is returned to last used app when more break windows are present

app/main.js
fdalvi: Agreed, `numberOfDisplays` is a good replacement!

app/main.js
fdalvi: Yes, I'll create a function `closeDisplays` which takes in a list of displays and closes all of them

fdalvi:
I will add in the changes, test and push tomorrow! I can only test my changes at work since I don't have a multi-display setup at home. Will confirm the window focus after a break as well.

fdalvi:
Changes are done and pushed. One question regarding locales, I had to add a `showOnAllScreens` entry in the en locale, should I add it in all other locales (with empty strings? Google translation?) as well?

hovancik:
Thanks, I'll test it.
I would add locales to every file (in English), so it's easier to keep track of what's missing.

fdalvi:
Done!

hovancik:
Thanks! 
Works fine on my Mac, will try on my linux machine. Focus seems to be brought back to last used app as well ;] 

Would you like to add yourself to README and make a change in CHANGELOG?

fdalvi:
Done! I updated `stretchly-settings-website-3.png` as well with the new setting - tried to maintain the same grayscale theme as the other screenshots. Let me know if anything else needs updating.

hovancik:
Works great ;]  so I guess should be fine on Windows as well. Thanks again for the contribution!

PR #210, opened 2018-02-14
Pause for 1.5 Hours

kavya-jain:
Added pause for 1.5 Hours-issue 89


hovancik:
Few things: 

 - welcome screen and pausing break reminders are 2 different things should be separate PRs as they are different things
- pausing break should be internationalised, not only in English. Ideally, it should be separate, testable class.

PR #217, opened 2018-02-18
Update zh-TW.json

wilicw:
modify some translation mistakes and some sentence


wilicw:
Traditional Chinese and Simplified Chinese are somewhat different in translation,I am not sure if my translation in Simplified Chinese is correct

wilicw:
I think there should be no other to be modified.

hovancik:
Ok. I'll merge it. Thanks!

PR #223, opened 2018-02-26
Feature: Auto Updater

tsah20:
Added auto updater as a dependency 
Added a basic  auto updater window to display the current version, latest version and download progress


package.json
```diff
@@ -120,6 +122,8 @@
   "dependencies": {
     "@paulcbetts/system-idle-time": "^1.0.4",
     "i18next": "^10.4.1",
+    "electron-log": "^2.2.14",
```
hovancik: Is this used anywhere? I don't see it defined

hovancik:
So I think we should talk about how is this supposed to work. But before that, let's not forget that we have basic update check already build in (please check how `function checkVersion ()` works). We need to either remove the older check, or combine them together. 

What do you think is the best course of action? And then, how do you think this should all work from user's point of view?

app/main.js
```diff
@@ -2,6 +2,7 @@
 const {app, BrowserWindow, Tray, Menu, ipcMain, shell, dialog, globalShortcut} = require('electron')
 const i18next = require('i18next')
 const Backend = require('i18next-node-fs-backend')
+const { autoUpdater } = require("electron-updater")
```
hovancik: we use [standardjs](https://standardjs.com/), so this should be `require('electron-updater')`. there are more issues, try running `npm run lint` to see them

app/main.js
```diff
@@ -155,14 +158,63 @@ function startProcessWin () {
...
+
+autoUpdater.on('update-not-available', (info) => {
+  sendStatusToUpdaterWindow('Update Not Available')
+})
+
+autoUpdater.on('error', (err) => {
+  sendStatusToUpdaterWindow('Error'+ err)
+})
+autoUpdater.on('download-progress', (progressObj) => {
+      let log_message = "Download speed: " + progressObj.bytesPerSecond;
```
hovancik: This will need to be translatable, but let's deal with it once the functionality is done.

app/main.js
tsah20: @hovancik  Yes we can combine the  two checks in  one.  But the `function checkVersion()` only notifies the user  of a new release and is internally using `VersionChecker` for checking updates. This can be easily handled through electrons `autoUpdater.checkForUpdatesAndNotify()` .  In addition to this what we can do is instead of redirecting it to the github release page (as in current system), we can redirect it to a prompt to update the newer version.  
[Prompt](https://lh3.googleusercontent.com/-WgkKa0lCy6M/Wppv13sWaqI/AAAAAAAAa2o/-l6ssKBeLQQZbzLhZrCusMbSZhTyQ5qZwCL0BGAYYCw/h1080/2018-03-03.png)

hovancik: Yes, that's where I was pointing. As you have it right now, `checkForAutoUpdate()` becomes available in menu only when old version checker finds new version. 

The picture looks good ;]

tsah20:
@hovancik  Yes we can combine the  two checks in  one.  But the `function checkVersion()` only notifies the user  of a new release and is internally using `VersionChecker` for checking updates. This can be easily handled through electrons `autoUpdater.checkForUpdatesAndNotify()` .  In addition to this what we can do is instead of redirecting it to the github release page (as in current system), we can redirect it to a prompt to update the newer version.  
[Prompt](https://lh3.googleusercontent.com/-WgkKa0lCy6M/Wppv13sWaqI/AAAAAAAAa2o/-l6ssKBeLQQZbzLhZrCusMbSZhTyQ5qZwCL0BGAYYCw/h1080/2018-03-03.png)

app/main.js
tsah20: :+1:  Will fix this

hovancik:
It looks good ;]

.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
```
hovancik: we should remove this file and add it do `.gitignore`

app/dev-app-update.yml
```diff
@@ -0,0 +1,3 @@
+owner: tsah20
```
hovancik: I see you added this file to the `.gitignore` 

Do they recommend to not commit it? 

If it's used for development, then maybe we should create `app/dev-app-update.yml.example` (so we don't need to google the format when we wanna use it for development ). What do you think?

app/main.js
```diff
@@ -22,6 +24,7 @@ let aboutWin = null
 let settingsWin = null
 let settings
 let isOnIndefinitePause
+let isNewVersionAvailable = null
```
hovancik: why did you created `isNewVersionAvailable`? We have global `isNewVersion` already. Do we need them both?

package.json
```diff
@@ -119,7 +119,12 @@
   },
   "dependencies": {
     "@paulcbetts/system-idle-time": "^1.0.4",
+    "ajv": "^6.2.1",
```
hovancik: Do we need this? I do not see it used anywhere.

package.json
```diff
@@ -119,7 +119,12 @@
...
+    "electron-builder": "^20.0.6",
```
hovancik: does `electron-builder` needs to be in app dependencies? we already have it in dev-dependencies and [documentation] (https://www.electron.build/auto-update) only talkes about `electron-updater`. 

Also: app dependencies needs to go to the `app/package.json` as we use 2 package.json structure. (https://www.electron.build/tutorials/two-package-structure)

app/utils/versionChecker.js
```diff
@@ -1,3 +1,5 @@
+const fetch = require('node-fetch');
```
hovancik: was there some issue with fetch, that you added is as a dependency?

Sammi-93: Yes, there was an issue. It was not able to find fetch. So, i added it.

tsah20: Yes there was an issue. It was not able find fetch. So i added it.

app/main.js
tsah20: I will check on it, if the global variable can be used.

package.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "stretchly",
-  "version": "0.15.0",
+  "version": "0.14.0",
   "description": "break time reminder app",
+  "main": "app/main.js",
```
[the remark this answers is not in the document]
tsah20: For debugging main app, I added vscode configuration. So it needed to have a refrence of main file.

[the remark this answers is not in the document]
tsah20: I will delete it if it's not needed.

package.json
tsah20: Yes we can remove it. But when we start our app with npm start, there are some warnings about dependencies being not resolved. I forcefully installed all the dependencies. I think its one one of the peerDependencies that needs to be added.

app/dev-app-update.yml
tsah20: Yes, It should be added in .gitignore. If we want to use updater in a dev mode,we need to add this file manually.
Yes we can add app/dev-app-update.yml.example for a reference.

package.json
tsah20: I added electron-builder for building and publishing my own releases. I will remove it.

.vscode/settings.json
tsah20: Yes, I will do it.

app/dev-app-update.yml
hovancik: Ok then ;]

package.json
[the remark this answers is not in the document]
hovancik: If it's for debugging we can keep it.

package.json
hovancik: What is the error? I am not getting any.

app/main.js
tsah20: The global variable is set to true within the first second of starting the app.So, the functionality as you suggested on gitter that a user will click on 'Check for update' and then the tray menu would be pushed with the 'Download latest version' would not be achievable because as soon as the app starts it calls for a function startProcessWin which checks for the new releases and notifies us. It also sets the global.shared.isNewVersion to true. So, what will happen is, we won't be able to see the 'Check for update' in the tray menu and the tray menu would be with the 'Download latest version' functionality when the app starts.

.vscode/settings.json
tsah20: Done

app/main.js
hovancik: I think that's ok. We were planning for automatic version check anyway, right? (once a day)

There is no point in users having to manually check for new version. They will only do it if they want to do it , and if our daily check has not been faster. The "Check for update" will be available most of the time (Until there is new version and our daily checker finds it).

package.json
hovancik: Does it cause some errors in app? if not, and as it is only WARN, I would ignore it. I'd rather have less dependencies.

PR #232, opened 2018-03-26
Wel page

kavya-jain:
Added a welcome page with translations and don't show again option


.history/app/css/app_20180305105330.css
```diff
@@ -0,0 +1,156 @@
+@font-face {
```
hovancik: you should add these files in gitignore.

kavya-jain: okay

app/locales/sk.json
```diff
@@ -109,5 +110,27 @@
   },
   "process": {
     "newVersionAvailable": "Je dostupná nová verzia!"
+  },
+  "welcome": {
+    "welcomeToStretchly": "Vitajte v Stretchly",
```
hovancik: Please do not use Google translate/or similar tool for translates. Those tools does not make correct translations. I would rather have English there and have it translated by someone native later.

app/main.js
```diff
@@ -651,7 +681,13 @@ ipcMain.on('save-setting', function (event, key, value) {
     breakPlanner.naturalBreaks(value)
   }
   settings.set(key, value)
-  settingsWin.webContents.send('renderSettings', settings.data)
+  //settingsWin.webContents.send('renderSettings', settings.data)
+  if (settingsWin) {
```
hovancik: Could you please check whether we have knowledge of what window is being used to send message (from event, or from electron api)?

I do not like doing `if` here. It's possible that someone could have settings and welcome window opened at same time and thus this would not work.

hovancik:
Hi,  I'm trying it right now and I think we are slowly getting there 🥇 

Few things: 
- let's not make changes to any translations files yet (google translate sucks and I would like to make final changes to texts)
- when making translates, we should not use strings from other files (eg. there is `<h2 data-i18next="main.selectLanguage"></h2>` in `index.html`). We should have only strings fixed to the window.
- I think we should merge the `index.html`and `welcome.html` files: 
    - keep the welcome.html, so it's more meaninful 
    - english/language from settings should be selected by default (one less user click)
    - also on settings3, let's not change that
    - add some short text like: 'welcome to stretchly, on next few pages we can have a look how to make the most of it' - I would look for some android app for inspiration (they usually have great intros). This might be instead of: "Perfect app that reminds you to take breaks while working on computer." 

I really like how you did the guide: the image and short text is great for showing feature 👍 . 

I'll try to make more comments in code. Do you think you could undone changes that were made on pictures and fonts? Not really sure why there are there and I think they should not be there: https://github.com/hovancik/stretchly/pull/232/files?short_path=3debddc#diff-3debddc669ee40c2f191f391bfecce1a

app/main.js
```diff
@@ -161,6 +165,32 @@ function startProcessWin () {
   })
 }
 
+
+function createWindow() {
```
hovancik: we should use some more meaningful name, eg `showWelcomeWindow`, so next contributors know what is happening here.

app/main.js
```diff
@@ -45,7 +49,7 @@ app.on('ready', startProcessWin)
 app.on('ready', loadSettings)
 app.on('ready', createTrayIcon)
 app.on('ready', startPowerMonitoring)
-
+app.on('ready', createWindow)
```
hovancik: I think that this is why I did  not see the Welcome page when I first tried it: 
- when app starts all these `app.on('ready'` are called, but as we are in nodejs, they are not called in sync way (one after another) 
- settings were not loaded yet, so `settings.get('showWelcomeWindow')` was something like `null` and I did not get window

Your solution works, but it's way too hacky: 
```
+  settings = new AppSettings(settingsFile)
+  if(settings.get('showWelcomeWindow') == null){
+    settings.set('showWelcomeWindow', true)
+  }
```

Instead we should create welcome window once settings are loaded (at the end of `function loadSettings ()`) or try to find if we can load settings before `app.on('ready', )`

app/locales/sk.json
kavya-jain: Okay. So should I replace all translations with English?

app/main.js
kavya-jain: I'll check

kavya-jain: Welcome window is the first window that opens up when the application starts running. So I don't think this code should create any problem.

hovancik: Let's keep for the end, when we have final wording.

app/locales/sk.json
hovancik: Let's keep for the end, when we have final wording.

kavya-jain: Okay

app/main.js
```diff
@@ -161,6 +165,34 @@ function startProcessWin () {
   })
 }
 
+
+function createWelcomeWindow() {
+  defSettings = new AppSettings(defaultSettings)
+  showWelcomeWindow = defSettings.get('showWelcomeWindow')
+  const dir = app.getPath('userData')
+  const settingsFile = `${dir}/config.json`
+  settings = new AppSettings(settingsFile)
```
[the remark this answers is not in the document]
kavya-jain: Okay. I've made the changes. Please review.

Issue #161, opened 2017-12-19
Snooze button


radu-gheorghe:
First of all, thanks a lot for doing Stretchly, the health industry should be grateful :)

Snooze is the feature I miss most from Workrave: 30s is usually too short of a notice for me to wrap up before taking a break (or maybe that can be made configurable?). Then when the break comes, I need to skip it and when I do, I forget about it.

If I could postpone the break for 5 minutes or so (configurable?), that would be awesome.

hovancik:
Hi, you can configure the reminder time by editing the config file: 
https://github.com/hovancik/stretchly#editing-breakmicrobreak-notification-interval 

The reason why it's not configurable is to make stretchly easy to understand. Settings are kept as small as possible (and are opinionated ) with option to edit them by hand in config file. 

About postponing: for me, I prefer not to have postponing feature, because then I would never take a break ;] I want stretchly to force me into break. But I will think about it

radu-gheorghe:
Thanks! I didn't know of the configuration file. I've edited that out to two minutes, hopefully that gives me time to wrap up and actually take a break :) For now, stretchly only really serves as a reminder to take breaks every now and then, I don't truly respect them. I think that if I had a postpone button, even if I'd use it repeatedly, it would be better than to skip breaks entirely (which is what I do now).

hovancik:
That makes sense. I can image having possibility to postpone break once.

hovancik:
@BullHorn7 you can pause the breaks and restart them when you are ready from tray menu

Issue #152, opened 2017-11-15
Idle time


jt007:
Great software! Here is an idea for some improvements.
I have noticed that your new version 12.0 reset the microbreak timer if I have a natural break from the computer of 5+ minutes. That's a great feature, but I have a suggestion for some improvements.
If I am out of office for less than 5 minutes, the microbreak may in this version start just a few seconds after I start using my computer. That’s a bit frustrating. In this case I have already had a microbreak (more than 20 sek break), so the microbreak timer should have been reset.
When reading on the computer, it sometimes may be no mouse movements for 1-2 minutes, so to make sure it actually has been a real natural break, it should require an idle time for 2 minutes to reset the microbreak timer. That’s my suggestion. What do you think?

hovancik:
Let's hear what other people think. I am open to the idea changing from 5 to 2 minutes... Not really sure if 2 or 5 is better... Or maybe have 5 min default and let user configure it.

jt007:
Thanks for your reply:)
I use the option “don’t let me skip microbreaks”, (I need it:)), so therefore it’s a bit frustrating to be forced to take another break just the moment I return to office after a break of up to 5 minutes.  
When I just return from the bathroom etc, I don’t need a 20 seconds break :)
Your proposal of letting users configure it would solve the problem:)

xim:
I'm for more configurability of idle time perception. A couple of descriptive configuration fields (e.g. pause timers after [2] minutes inactivity, reset microbreak timer after [5] minutes inactivity, reset normal break timer after [15] minutes inactivity) also automatically tells the user exactly how the "monitor system idle time for natural breaks" works.

kostadinnm:
Hi there!
Kudos for the great app!
I'd like to add a suggestion for the idling monitor - could we have some kind of monitoring the system usage(say the cpu usage) in order to have more accuracy on whether a user is sitting in front. Another thing will be whether the image on the monitor is changing.
While typing this, I realize that the best thing would be having a dedicated sensor(and an accompanying api) like this on some samsung smartphones:
![smartstay](https://user-images.githubusercontent.com/18436405/36390554-2655cdd8-15ac-11e8-8bc5-5154907ed908.jpg)
Anyways, I guess a dummy implementation using the resources could work for some of us(e.g. my little cousin watching youtube) for a while.
Cheers,
Kostadin

lundibundi:
[...]

PR #233, opened 2018-03-27
Add ability to postpone breaks

lundibundi:
* allow postponing breaks in non-strict mode (break/microbreak Postpone)
* adjustable postpone time (break/microbreak PostponeTime)
* after a configurable number of postpones show skip button instead of postpone
  (0 for infinite postpone) (break/microbreak PostponesLimit)

Also currently there is no option in Settings for Postpone because I'm not sure for which ones to put there as I would like to get your opinion on this.

This is my first time working with electron so I hope I got everything right.

UPD: I've got an idea to use negative postponeLimit numbers to note that break must become strict after that amount of postpones, what do you think about it? (or use positive for this purpose and negative when we allow skipping the break) Or just add another option for such behavior.

fixes: https://github.com/hovancik/stretchly/issues/161


lundibundi:
CI failure on mac seems unrelated to me.

hovancik:
Hi,  I'm going for holidays now, so not sure if I manage to take a look before I leave, so please be patient. I'll be coming back at the end of the April.

lundibundi:
@hovancik Have a good holiday =). I'd also like to get your opinion on https://github.com/hovancik/stretchly/issues/152#issuecomment-376507773 as I wanted to implement it this way before you go if possible.

hovancik:
@lundibundi I'm still not decided how to do about idle time detection, so please do not implement anything, yet.

hovancik:
Hey, so I am trying it right now. 

My first thoughts:  
- I would only allow to edit settings by manually editing config file (I want to keep settings simple, and I do not think we need  this settings on Settings page - the most of people I talked to never change any settings)
- skipping should be enabled by default
- number of allowed skips should be one (yeah, let's be harsh ;] )

In the beginning I didn't like the fact that you are changing "skip" for "postpone" - reason being I thought user should have both options. And also I wanted to recommend another solution: user can postpone break by clicking on break notification. But I am beginning to like your solution: it's more educational  and breaks should be more often if users postpone once than skipping. I will think more about it. Also not sure what about times: 2 and 5 minutes. Maybe both as 2 minutes? 

Also: in future, I want to have idle time detection ala: when break is about to start, we will wait some time until user is idle and then start the break. So this is also something to keep in mind. 

For your UPD question:  
I've implemented (long long time ago) https://github.com/hovancik/stretchly/tree/feature/dangerous-breaks which works kinda like this: it counts number of skipped breaks and increases a variable, when break is not skipped it decreases the number. Based on the value of variable, it makes break window look like when you are dying in game (red spots): higher value, more red. I had problems with CSS and retina displays so I never fixed it ;/ 

So anyway, right now I would not implement "strict mode enforcement" - one change at time ;] but in future there could be some kind of strict mode via what you have described or style I described.  

An one last thing, I have not read your code yet, will find time after I come back from holidays.

hovancik:
One problem I noticed: 
- break starts as normal
- user can postpone
- let's say it last 20 seconds till the end of break 
- user does `cmd+x`  

Should we skip break of postpone? I think that we should skip. What do you think?

lundibundi:
@hovancik As for the first message

> I would only allow to edit settings by manually editing config file (I want to keep settings simple, and I do not think we need this settings on Settings page - the most of people I talked to never change any settings)

I'm completely fine with only having those options in config file, I'm used to editing it anyway).

> skipping should be enabled by default
number of allowed skips should be one (yeah, let's be harsh ;] )

That's fine, the only reason I set it to false was to preserve current behavior.

> Also not sure what about times: 2 and 5 minutes. Maybe both as 2 minutes?

lundibundi:
@hovancik ping.

hovancik:
Hi, sorry for late reply. I am super busy with another PR but no worries, I did not forget ;]

hovancik:
I've just read related issue and our previous conversation. Uff ;] 

I'm still thinking about what do to with skip/postpone when the break is let's say at 80%. I'm beginning to think that ability to postpone should be available only first 10/20% of break time.. 

Let's say I'm 2 minutes into 5 minute break, I guess we can safely say that one does not want to postpone but skip? I am not sure...

lundibundi:
This is truly a hard thing to define, but I think 10% should be okay (or maybe 20% as default to have it at 1 minute for default 5 min long break), as you usually don't need to postpone a break if half of it has already passed. I think we can also add a configuration option for this.

Though I think this should not be available for short breaks because they are short enough already.

lundibundi:
Ok, so there is the case when you can pospone but you cannot skip (strictMode: true, breakPostponable: true), so what should we do in that case if the `breakPostponableDurationPercent` (what do you think about the name btw?) is reached?
I suggest that in this case we should allow to postpone for the whole duration of the event, so that later when user reaches postpone limit he will not have any option available (neither postpone nor skip). and will have to take a break (which is what strict for).

hovancik:
The name is better than mine, I like it ;] I will try it out :+1:

hovancik:
Sorry for delay, I would like to work on this now. I'll try to merge master in.

hovancik:
I've created new branch, https://github.com/hovancik/stretchly/compare/master...feature/postpone 
Will need to test if merge was OK

PR #260, opened 2018-07-02
Fix manual breaks upon suspend/resume

### LINKED ISSUE #259: resumes from manual pause upon system resume
### Prerequisites

- [x] I'm using version: master at c9d960eeb801921820ad36435717e7c4cc204936 with https://github.com/hovancik/stretchly/pull/233 merged in.
- [x] I'm using version: master at c9d960eeb801921820ad36435717e7c4cc204936.

### Description

Stretchly resumes manual pause upon resume from suspend.

Operating system: Arch linux

### Steps to Reproduce

1. Start app
2. Select pause for 1 hour in menu
3. Suspend the pc
4. Resume from suspend

**Expected behavior:** Stretchly should stay paused

**Actual behavior:** Stretchly is no longer paused

**Reproduces how often:** always

Based on code inspection and in particular `keepAfterPowerResume` this seems to be the intented behavior but this seems wrong to me, so I suggest to change it to preserve any user breakand not only indefinite break. What do you think?

Also as there is currently no way to correct break time after resume from suspend (for example user start 1 hour pause, then suspends the pc for 2 hours and if not for the aforementioned behavior stretchly would still be paused) I suggest to add such correction mechanism to power-state listemers and breakPlanner. This may be done with simple system suspend time logging and then computing passed time upon resume to use for correction.

lundibundi:
* Only resume automatic for-suspend break
* If was on break then correct break time based on suspend duration
* Remove unnecessary logging

Issue: https://github.com/hovancik/stretchly/issues/259

closes #259 

### Requirements

- [x]  all new code has tests to ensure against regressions. (most does excluding actual suspend/resume testing)
- [x] `npm run lint` reports no offenses. (fails for eslint-chai problems)
- [x] `npm run test` is error-free. (fails on non related test)

### Description of the Change

This PR makes so that user breaks are preserved after system suspend/resume and corrects the break time accordingly.
Based on electron issues (see https://github.com/hovancik/stretchly/issues/259#issuecomment-401622387) it needs to be tested on win and mac devices more.
Also I'll add tests and readme/changelog changes later.

### Verification Process

- [x] Verified that it preserves user pause and corrects pause time (via logging pause time and suspending the pc)
- [x] Verified that it pauses/resumes breaks for suspend when the was no user pause (previous behavior) (via logging pause time and suspending the pc)

### Other

Why don't we use node-version list of all supported versions? It seemed to work just fine on node 10.5.0.

### DIFF
--- a/app/breaksPlanner.js
+++ b/app/breaksPlanner.js
@@ -135,8 +135,8 @@ class BreaksPlanner extends EventEmitter {
     this.nextBreak()
   }
 
-  restart (correctByMillis = 0) {
-    if (this.scheduler) this.scheduler.reload(correctByMillis)
+  correctScheduler () {
+    if (this.scheduler) this.scheduler.correct()
   }
 
   reset () {
--- a/app/breaksPlanner.js
+++ b/app/breaksPlanner.js
@@ -135,9 +135,8 @@ class BreaksPlanner extends EventEmitter {
     this.nextBreak()
   }
 
-  correctBy (millis) {
-    if (!this.scheduler) return
-    this.scheduler.reload(millis)
+  restart (correctByMillis = 0) {
+    if (this.scheduler) this.scheduler.reload(correctByMillis)
   }
 
   reset () {
--- a/app/breaksPlanner.js
+++ b/app/breaksPlanner.js
@@ -135,6 +135,11 @@ class BreaksPlanner extends EventEmitter {
     this.nextBreak()
   }
 
+  correctBy (millis) {
+    if (!this.scheduler) return
+    this.scheduler.reload(millis)
+  }
+
   reset () {
     this.clear()
     this.resume()
--- a/app/main.js
+++ b/app/main.js
@@ -82,9 +82,7 @@ i18next.on('languageChanged', function (lng) {
 
 function startPowerMonitoring () {
   const electron = require('electron')
-  let suspendStart = Date.now()
   electron.powerMonitor.on('suspend', () => {
-    suspendStart = Date.now()
     if (!breakPlanner.isPaused) {
       pausedForSuspend = true
       pauseBreaks(1)
@@ -95,8 +93,8 @@ function startPowerMonitoring () {
       pausedForSuspend = false
       resumeBreaks()
     } else if (breakPlanner.isPaused) {
-      const suspendDuration = Date.now() - suspendStart
-      breakPlanner.correctBy(suspendDuration)
+      // restart the planner to correct for time spent in suspend
+      breakPlanner.restart()
     }
   })
 }
--- a/app/main.js
+++ b/app/main.js
@@ -83,18 +83,20 @@ i18next.on('languageChanged', function (lng) {
 function startPowerMonitoring () {
   const electron = require('electron')
   electron.powerMonitor.on('suspend', () => {
+    console.log('The system is going to sleep')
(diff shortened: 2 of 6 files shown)


hovancik:
I will check code later and test it but to answer: `Why don't we use node-version list of all supported versions? It seemed to work just fine on node 10.5.0.`

I am copying version that electron uses:  they have it specified at https://electronjs.org/

I guess we could use any version, but just to be sure I am always using the same as they do.

app/main.js
```diff
@@ -82,13 +82,22 @@ i18next.on('languageChanged', function (lng) {
 
 function startPowerMonitoring () {
   const electron = require('electron')
+  let suspendStart = Date.now()
```
hovancik: this should only be `let suspendStart`, right?

app/breaksPlanner.js
```diff
@@ -135,6 +135,11 @@ class BreaksPlanner extends EventEmitter {
     this.nextBreak()
   }
 
+  correctBy (millis) {
+    if (!this.scheduler) return
+    this.scheduler.reload(millis)
```
hovancik: I am not sure if I would correct the timer here, or inside scheduler. But I guess you are right, scheduler should be able to reschedule itself. So maybe rename the method to `reschedule`? Tell me what you think ;]

lundibundi: My reasoning was as follows: the scheduler knows about it's start/delay times so it shouldn't leak that information to the breakPlanner but at the same time it should allow to correct time/reschedule itself as a black box.

It's really kind of hard to come up with a decent name here. I don't think `reload` is a good choice, but I didn't want to have `scheduler.reschedule()` thingie, as it's sort of repetitive. Though both of them are not good enough imo, so I'm fine with using `reschedule` if you want? =)

lundibundi:
@hovancik Oh, my mistake, I corrected the timeLeft, but it's not needed, it's already correct, because it's absolute. So basically we only need to restart the scheduler's timer with newly computed timeLeft because setTimeout is actually the problem and not the timeLeft (setTimeout doesn't count the time while in sleep, hence the problem).

I guess I'll left the method as is, maybe we will actually need it, but for now we only actually need to restart the timer itself.

hovancik:
Ah, it's a bit confusing ;] From what I gathered we only need to change the timer (without changing `Scheduler#now`) so it's called in `Scheduler#timeLeft`, right? 

So I guess we should simplify it. 

(0. let's keep the log messages about start/end of suspend, they are useful) 
1. when we go to sleep, we only go infinite when we are not on pause (as you did)
2. otherwise we let the pause (as you did)
3. when we come back, we clear the timer and start it with time left


The code with correcting value is OK, I guess, but I would not keep it. If we do not use it now, we probably wont be using it later. And in 2 months I wont be sure if I'm using it at all and I will be afraid to remove it. 

How about having `BreakPlanner.correctScheduler` and `Scheduler#correct` ? I guess that's what we are doing, correcting the scheduler. I have no problem with longer names, as far as they are descriptive so I guess we could also have `BreakPlanner.correctSchedulerTimer` and  `Scheduler#correctTimer`

Then, I think, we need to test what is gonna happen if we pause for hour but sleep longer.

lundibundi:
Okay.
As for the last question, I manually check the time to avoid additional delay of timer, but I think I'll just reschedule the timer, as negative delay will be set to 1 so it should be okay.

Issue #243, opened 2018-04-20
Pause until tomorrow


nfelger:
This is a feature suggestion.

I like using Stretchly all when I'm working, but I use the same laptop for work and leisure and I often find myself in a position where I'm watching videos or doing other things with my laptop in the evening. In those situations, I don't always like having the breaks.

What would be nice for that is a menu option to pause Stretchly until the next morning (e.g. 6am local time), similar to the "Disable until sunrise" feature in f.lux (https://justgetflux.com/).

What do you think about this?

hovancik:
I like it ;]  
The only question is what would "tomorrow" mean.  I guess 6am could be ok

lundibundi:
I'll work on it. I think I'll implement 2 versions
* manual hour specification
* specification of coordinates so we will be able to calculate the actual sunrise
@hovancik what do you think?

lundibundi:
Oh, that's good, but imo it's easier to just push `until morning` than to manually specify time (if I understand correctly how custom pause will work) because that's much more common use case, for example when I'm finished with work for today and want to watch films etc.

Location thing is completely optional and available only through manual edit of 'config.json' so this shouldn't be that big of a deal, even more so this only needs approximate location so city or even country will work just fine.

PR #262, opened 2018-07-04
Implement pause until morning

lundibundi:
* allow users to pause until specified morningHour
* allow to pause until sunrise (morningHour = sunrise and posLatitude
  and posLongitude have to be specified)

Issue: https://github.com/hovancik/stretchly/issues/243

### Requirements

- [ ]  all new code has tests to ensure against regressions. (cannot really test UI stuff)
- [x] `npm run lint` reports no offenses. (only irrelevant)
- [x] `npm run test` is error-free. (only irrelevant)
        Changelog updated, not really sure what to put in Readme.

### Description of the Change
Allows users to pause until specified `morningHour` in settings or until 'sunrise' which is calculated based on location provided by user.

### Verification Process

- [x] Only tested manually that it correctly sets pause time

### Other

It depends on https://github.com/hovancik/stretchly/pull/260 for correctness because this pause will be reset upon system suspend with current behavior.


lundibundi:
@aleuca Oh, you talk about error case if loadMorningTime, this shouldn't happen unless user goes to `config.json` and manually sets something like `"morningHour: "invalid"` so I guess it should be kind of okay. Otherwise, we may reuse defaultSettings.js value there.

Also, it won't be a problem if you merge the PR for auto update before this, I'll just rebase =).

package.json
```diff
@@ -121,6 +121,7 @@
   "dependencies": {
     "@paulcbetts/system-idle-time": "1.0.4",
     "i18next": "11.2.3",
-    "i18next-node-fs-backend": "1.0.0"
+    "i18next-node-fs-backend": "1.0.0",
+    "suncalc": "^1.8.0"
```
hovancik: this will need to be in `app/package.json` as well so it get's packaged to the installers and so on 

cc @aleuca @Sveijk

app/main.js
```diff
@@ -573,6 +574,17 @@ function getTrayMenu () {
...
+        }, {
+          label: i18next.t('main.untilMorning'),
+          click: function () {
+            const morningTimes = loadMorningTime()
+            if (morningTimes === null) {
+              console.log('Failed to load morning time, cannot pause')
+              return
+            }
+            const untilMorning = Utils.millisUntil(...morningTimes)
+            pauseBreaks(untilMorning)
```
hovancik: can you create a class for this in `/utils`? so in the end we would only call `pauseBreaks(Something.untilMorning())`or so

main.js is way too big already

app/locales/en.json
```diff
@@ -16,6 +16,7 @@
     "forHour": "for an hour",
     "for2Hours": "for 2 hours",
     "for5Hours": "for 5 hours",
+    "untilMorning": "until morning",
```
hovancik: we will need to copy this to all localization files

lundibundi: I thought that english translation will be used by default if there is no translation in the selected language (no entry in the file)?

app/main.js
lundibundi: I'd suggest to make it just a file with a few functions that will return morning time, there is no need in a class imo.

lundibundi:
@hovancik I'm fine with it, but what would be the best way for them to make changes. Should I give them access to my fork so they will be able to commit to this branch?

lundibundi:
@hovancik that's just a fork anyways) I've added @Sveijk and @aleuca as collaborators, thay should get an email and after accepting be able to push to this branch. Though don't forget that this have to be rebased first.

lundibundi:
@Sveijk and @aleuca you seem to have made a few mistakes mostly related to git, so I'd like to clarify them and offer a solution. 


Though don't worry this happens with everyone, git takes time to master :).


* You have merged master branch of **my** fork which is actually a PR for break-postpone, it shouldn't be here. I initially started it on master and couldn't move afterwards. The master from @hovanchik is actually 'umaster' branch. 
* There are some unmerged changes in default-settings.js, you don't want to commit those
* My commits are somehow twice in this branch now (see commits after your first merge commit), this shouldn't happen, idk actually how it happened.
* Two latest commits don't have an author, the one who committed it probably doesn't have email/name set up in git or GitHub (this is probably @Sveijk). Check out [this](https://help.github.com/articles/setting-your-username-in-git/) and [this](https://help.github.com/articles/setting-your-commit-email-address-in-git/) GitHub tutorials.


I can't offer any other solution but to git hard reset (**this will remove any changes until that commit** make sure to save you work (see later)) to my latest commit (before your merge) and redo it. First merge/rebase (I still suggest a rebase as it would make it cleaner) an then your changes. There are a few ways in git to avoid re-coding everything again (e.g. stash, save to other branch then cherry-pick (probably best here) after clean up of this branch, interactive rebase to real master while removing commits). But all of them require some more advanced knowledge of git, so unless you want/have time to learn it now I'd suggest you instead just copy files you have changed to a separate (non this project) directory and after hard reset and rebase copy them back with appropriate changes, this should be easy and fast for now.


Good luck and have fun learning. :wink:

Sveijk:
@lundibundi @aleuca Last issue in the row you mentioned, I solved. There was a different git user.name and user.email I used indeed. So I made a local git config for this repository as well. It has been my work-account for sure.

- We didn’t merge into master of Stretchly yet. We thought it best that you would merge your repo into master. Since, there’s a bit more risky business involved in that step ;-). 

- It’s voodoo magic, I don’t understand how your commits are now twice in this branch.
I guess this 'problem' will also be solved when `git reset --hard` is done. Because that would rewrite history, all newer commits will be lost with this command.

Another question. Could it be that umaster is local on your working directory? Because when I try to fetch it, it returns a fatal error. 

Thank you for your feedback! We're eager to learn and the best way is to learn from what we did wrong I guess ;-)

lundibundi:
@Sveijk oops, my bad, I've uploaded `umaster` branch. Also you can always use `git pull --rebase upsteam master` to get @hovancik 's master (assuming you have set up upsteam remote, check out [this](https://help.github.com/articles/fork-a-repo/) or you can just try the command as it will fail if you don't have upstream set up =)).

> We didn’t merge into master of Stretchly yet. We thought it best that you would merge your repo into master. Since, there’s a bit more risky business involved in that step ;-).

I meant to merge Stretchly master into this branch. I can do that but as I mentioned this will need a hard reset probably so this will remove your changes and I don't want that result, hence I suggested that you save your changes -> hard reset -> reapply changes

> It’s voodoo magic, I don’t understand how your commits are now twice in this branch.

This is git, there are ways to do anything) But yes hard reset will help but make sure to save your changes somewhere.

Your coarch is correct that it is better to first merge/rebase and only then apply changes as this will minimise the amount of conflicts you have to solve. The only thing is that you forgot to check that you've successfully rebased/merged. Hence now this problematic rebase/merge propagated into the next commits so this has become much harder to fix without hard reset.

Yep, it's fine to make mistakes, don't worry :wink:.

PR #269, opened 2018-07-08
Update package.json versions and app tests (in progress)

### LINKED ISSUE #267: Lock all npm dependencies to exact versions instead of auto-update
Currently all dependencies are on auto-update. We should lock in the exact versions to avoid breaks in the future when dependencies come out with new versions. From now on when we run `npm i dependency_name --save --save-exact` to always lock in the current version.

### LINKED ISSUE #268: "on app start" test needs to be updated
### Prerequisites

- [x] I'm using latest version: v 0.17.0 

### Description

"on app start" test checks for 2 open windows.

### Steps to Reproduce

1. run `npm test`

**Expected behavior:** Expect test to be checking for 2 process windows on first run and 1 process window on open if the user is running app not for the first time.

**Actual behavior:**  test fails and test checks for 2 windows instead of checking on condition whether or not the user is running the app for the first time

**Reproduces how often:**  100%

aleuca:
Issue: #267 #268 

### Description of the Change

- Changed package.json to not autoupdate npm packages. From now on any "npm i" command will install the exact versions and will not update package-lock.json

-Updated tests to check for 1 window on app open and 2 windows on app open if user runs app for the first time

### Verification Process

- run `npm test`
- run `npm i`

closes #267, closes #268


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Fixed & Added
```
hovancik: I put only app features to changelog, so no need to write that we added test and created helper

CHANGELOG.md
```diff
@@ -19,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Czech translations updated
 - Slovak translations updated
 
+
```
hovancik: let's get rid of empty line ;]

test/helper.js
```diff
@@ -0,0 +1,34 @@
+let Application = require('spectron').Application
```
hovancik: let's call this file `modifySettings`or `modifySettingsHelper` ? `helper` is not very descriptive

PR #276, opened 2018-07-18
Do not have breaks in DND mode (win, mac)

### LINKED ISSUE #240: Stretchly not paused automatically during DND/presentation mode on OSX
### Prerequisites

I'm using latest version: v0.16.0. I am also running stretchly in full screen mode.

### Description

When enabling do not disturb mode (or when done automatically during presentation mode) stretchly still works. I've had the messages  pop up when I am trying to present things to other people which is really annoying.

I could switch it off, but I forget, and worst still the DND mode turns off the stretchly notifications (as expected) which let you know it will happen shortly.

Operating system: OSX high sierra 10.13.2

### Steps to Reproduce

1. Go to notification centre and turn DND on
2. Wait for microbreak or break and stretchly will start

**Expected behaviour:** When in DND stretchly should pause for the duration of DND

**Actual behaviour:** Stretchly interrupts presentations and causes some bother.

**Reproduces how often:** 100% of the time

Many thanks for reading.

Sveijk:
Issue: #240 

### Requirements

### Description of the Change

For Mac users when in Do not disturb mode, the breaks won't run. Notifications will not be shown by Stretchly.

### Verification Process

If DND mode is on, breaks won't run. If the DND mode is off the breaks and notifications will be shown to the user.

### Other information
closes #240


Sveijk:
@hovancik the DND has been updated. Now the tray menu displays when the user is in Do Not Disturb mode. The Menu will be updated at the same interval as a break interval. This update is done inside the microbreak notification and the break notification function.

app/about.js
```diff
@@ -49,6 +49,8 @@ ipcRenderer.on('debugInfo', (event, reference, timeleft, settingsfile) => {
     timeleftElement.innerHTML = timeleft
     let settingsfileElement = document.getElementById('settingsfile')
     settingsfileElement.innerHTML = settingsfile
+    let dndElement = document.getElementById('dndElement')
+    dndElement.innerHTML = getDnD
```
hovancik: Think about better name for `getDND`. That's how you could name the function to get DND state. Check others ie `timeleft` or `settingsfile`: it clearly says what it is.

app/locales/bg.json
```diff
@@ -33,7 +33,8 @@
     "microbreak": "микропочивка",
     "break": "почивка",
     "nextBreakFollowing": "\nСледващата почивка ще бъде след {{count}} микропочивка",
-    "nextBreakFollowing_plural": "\nСледващата почивка ще бъде след {{count}} микропочивки"
+    "nextBreakFollowing_plural": "\nСледващата почивка ще бъде след {{count}} микропочивки",
+    "notificationStatus": "In do not disturb mode"
```
hovancik: if you check the tray message, you will notice that `In do not disturb mode` is not on right position (it stays in first line: `stretchly - break time reminder appIn do not disturb mode`). You can new line with `\n`

app/main.js
```diff
@@ -1,7 +1,8 @@
 // process.on('uncaughtException', (...args) => console.error(...args))
-const {app, BrowserWindow, Tray, Menu, ipcMain, shell, dialog, globalShortcut} = require('electron')
+const { app, BrowserWindow, Tray, Menu, ipcMain, shell, dialog, globalShortcut } = require('electron')
```
hovancik: so somewhere we use `{ xxx` and somewhere `{xxx`. We should decide which one we wanna have. Are there some reason to have a space there?

hovancik: I guess the same would go for `()`

hovancik: I noticed that for `()` we use no space

app/locales/cs.json
```diff
@@ -34,7 +34,8 @@
     "break": "přestávky",
     "nextBreakFollowing_0": "\nDalší přestávka po {{count}} mikropřestávce",
     "nextBreakFollowing_1": "\nDalší přestávka po {{count}} mikropřestávkách",
-    "nextBreakFollowing_2": "\nDalší přestávka po {{count}} mikropřestávkách"
+    "nextBreakFollowing_2": "\nDalší přestávka po {{count}} mikropřestávkách",
+    "notificationStatus": "In do not disturb mode"
```
hovancik: [Apple](https://support.apple.com/en-us/HT204321) calls this `Do Not Disturb`so we should do the same

package.json
```diff
@@ -122,7 +122,9 @@
     ]
   },
   "dependencies": {
+    "@meetfranz/electron-notification-state": "^1.0.0",
     "@paulcbetts/system-idle-time": "^1.0.4",
+    "@sindresorhus/do-not-disturb": "^0.2.0",
```
hovancik: we are not using this anymore, right?

app/main.js
Sveijk: Must have been something that happened because of space settings in my code editor, I'll fix them back to no spaces.

hovancik: I don't know which one is correct ;] Probably different editores/people use different settings

app/locales/bg.json
hovancik: Or maybe add new line in code where we create tooltip message?

package.json
Sveijk: Yes we don't use it, I deprecated it :-)

PR #281, opened 2018-07-30
Set monochrome tray icon as default

### LINKED ISSUE #253: Monochrome icons for all platforms
Can you enable the monochrome icons for all platforms (not just OS X). For example both KDE and Windows 10 use white monochrome icons.

hovancik:
- user can change this in settings

Issue: closes #253 

### Requirements

### Description of the Change

- default tray icon will be monochrome
- user can change that in settings 3 

### Verification Process

- default settings are applied on first run
- changing icon in settings changes the icon
- resetting settings to defaults works


hovancik:
Yeah, I need to look into icons in the future. They are tricky and works different on different platforms/resolutions

PR #285, opened 2018-08-11
add time of next break to tray menu, fix english error in locales

### LINKED ISSUE #148: Feature request | show time remaining for next break in menubar
It would be cool, if the time remaining for the next break can be shown in the menu bar.

aleuca:
Issue: #148 

### Requirements

### Description of the Change
Add a tray menu item that says when the next break is going to happen, hour:minutes.

### Other information

closes #148


app/main.js
```diff
@@ -693,6 +702,7 @@ function updateToolTip () {
             statusMessage += i18next.t('main.nextBreakFollowing', {'count': breakInterval - breakNumber})
           }
         }
+        return breakType
```
hovancik: why do we return `breakType` here? it's causing tootlip not to update

app/main.js
```diff
@@ -479,6 +469,11 @@ function saveDefaultsFor (array, next) {
 
 function getTrayMenu () {
   let trayMenu = []
+  let timeLeft = breakPlanner.scheduler.timeLeft
```
hovancik: can we moved this `minutes` creation to utils or somewhere? we could also test it out there with a simple test ;]

Issue #57, opened 2017-01-24
Feedback


hovancik:
Hi, I got some feedback from https://github.com/sindresorhus/awesome-electron/pull/420#issuecomment-274509218

What do you think? (if anyone here ;]). I will implement some stuff, but I don't plan to work on plaform-specific stuff.  

```
Some feedback. For macOS, but might also apply to other platforms:

App name should be captialized. stretchly => Stretchly.
I would recommend reading https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/ Your app doesn't feel very native to the platform.
Menubar icons should not be colored.
Quit should be Quit Stretchly and also have a shortcut to quit.
All the menu items should be titleized.
Settings => Preferences...
About should be at the bottom above Quit.
There should be a separator above Settings.
The About window is way too big, it should not be resizable, and the links there are not styled.
Don't use a notification to inform me that Settings are saved when changed. That's just annoying.
The Settings window should not be resizable.
A lot of the text is too hard to read because it its thinness and bad contrast.
The "buttons" in the app are not clear enough. They're just normal text without even changing the mouse pointer.
Drop the splash screen. That's just annoying.
Since you already have a settings window, you should move "start at login" there, to reduce stuff in the menu.
If I click "Pause - for two hours", it should show me how much time left there is of the pause.
```

niladam:
Most of the feedback you have there is very good. I'm opening my own issue now too :)

hovancik:
so some stuff was implemented. closing for now, people are welcome to open their own issues.

PR #299, opened 2018-09-22
Fix text readability, change from hairline/light/thin to regular/bold

jancborchardt:
Issue: #57

Specifically the part

> A lot of the text is too hard to read because it its thinness and bad contrast.

### Requirements

   - I didn’t specifically open an issue, but the existing #57 already mentioned the problem.
   - Wasn’t sure how to best test loop through all the breaks and microbreaks. The indicator icon to get to the settings was not shown in development (am on Ubuntu).
   - Didn’t take new screenshots of all the pages.

### Description of the Change

The hairline, thin and light font weights make the text very difficult to read and result in subpar contrast on basically on the colors. You should be able to read the break suggestions quickly so you can perform them, and not harm your eyes further.

That’s why the font weights were replaced with higher weights from the same font, Lato:
- Hairline to Regular
- Thin and Light to Bold (also to not have too many different weights which are almost the same anyway)

I downloaded Lato from Google Fonts and it only included the ttf. If all the other formats are indeed needed, let me know where you downloaded them from and I’ll add the files for Regular and Bold.

### Verification Process

I checked out the initial run window and it looked good. Can’t make it show up again though to take screenshots, and can’t loop through the break suggestions or show the settings window as mentioned above. :\ Let me know how it works and I’ll continue with that. :)


jancborchardt:
The Appveyor failure seems to be unrelated to this change as only CSS and font files were touched? https://ci.appveyor.com/project/hovancik/stretchly/build/1.0.1021/job/qiqjt37x16ai3xrs

hovancik:
Yeah, appveyor seems to fails tests on random lately. 

You might need https://github.com/hovancik/stretchly#linux-note to get tray working correctly on Linux. 

Here is what I am getting on Linux with current version: 

screenshot_2018-09-23_11-04-21

and on mac: 

stretchly-break

and with your change it looks like this (on Linux):

screenshot_2018-09-23_11-13-27
 
so this is something platform specific... Maybe some font related CSS? I'm not sure what mac/linux/windows are doing differently when rendering fonts, but I think that's where solution might be. What do you think?

hovancik:
this is mac: 
screen shot 2018-09-23 at 16 10 09

jancborchardt:
I would definitely say it’s way easier to read. Yes, on macOS it seems less bad in the current version than on Linux (which I am using), but still thin fonts are not ideal.

I would actually do additional things like:
- Bold the heading (regular for body text) for better separation – **this relates a lot with this pull request**
- Possibly cut down the text and improve wording of each break so there’s less to read
- Making the body text left-aligned for even more readability
- Introduce a symbol (maybe emoji) to show for each activity to even more quickly see what it’s about

And all of these would change the look quite a lot – but it’s best to do stuff step-by-step so it’s easier to review and discuss. So yes, I definitely think that this pull request is a step in the right direction, but it’s only the first one.

What do you think?

hovancik:
I've created #304 and tested on win, amc, debian. Thanks for bringing this issue forward ;]

PR #312, opened 2018-10-22
[It] Italian translation of tutorial and refinements

cgand:
[It] Italian translation of tutorial and consistent refinements of interface.


cgand:
Thanks for keep the Italian translation updated.

PR #321, opened 2018-10-27
Single page settings

Jackymancs4:
### Requirements

### Description of the Change

I unified all settings pages under a single one, with one main script handling it.
The internal semantic division is kept mostly unchanged, so settings are still treated in three separated groups (settings, settings2, settings3).

To keep the very same layout, I implemented a purely CSS carousel following this [tutorial](https://codeburst.io/how-to-pure-css-carousel-ce1a8cb231c8).

Now scrolling through the settings panes is smooth thanks to the customizable CSS transition function, and doesn't require the new page load lag anymore.

### Other information

|Before|After|
|--|--|
|schermata 2018-10-27 alle 16 54 40|schermata 2018-10-27 alle 16 54 12|

If required I can post a recording of the changes applied.
If and when merged, the next step could be refactoring the slighlty big settings file.
Sorry for not opening a dedicated issue before, it's really been a quick idea that worked out good.


hovancik:
Thanks, I will test it around ;]

hovancik:
I like it, but there are some problems I have found: 

screenshot_2018-11-03_17-22-44 

on first 2 pages I have some scrollbar at the bottom, but not on 3th and it makes page jump a bit 

Also when you change the language on 3th page it jumps to the 1st page

Jackymancs4:
So, two bugs:

1. - [ ] on first 2 pages I have some scrollbar at the bottom, but not on 3th and it makes page jump a bit
2. - [ ] when you change the language on 3th page it jumps to the 1st page

I'm able to replicate both.

1. easy to fix, patch on its way
2. this one seems tricky. 

I have not investigated yet how you (or i18next) propagate the new language data, but it seems to me like a page refresh, in which case I'm not sure how to retain the panes status. If this is a strict requirement I'll need a couple days to think about it.

hovancik:
Yeah, I think when we change the language, we wanna stay at the same page. You might be able to do that with # link. 

How difficult would be to add one more page to settings?

hovancik:
Reworking the settings based on designs from designer so closing this one.

Issue #327, opened 2018-10-30
[Request] Disable notification sound before alert


asterite:
### Prerequisites

- [x] I'm using latest version: v0.18.0

### Description

[Description of the issue]

Operating system: macOS High Sierra 10.13.6

### Steps to Reproduce

1. Click on the tray icon
2. Click on "Settings"
3. Click on the right arrow to advance to the next page
4. In "audio scheme" choose "silence"

**Expected behavior:** If I choose "silence" I expect no sounds at all coming from the application

**Actual behavior:** A notification sound appears before break and microbreak, when the notification pops up (for example the one that says "next microbreak in 10 seconds".

**Reproduces how often:** All the time

### Additional Information

I guess the "audio scheme" refers to the sound that appears when stretchly becomes active or deactive?

I personally use stretchly without sounds and just the notification that appears on screen 10~30 seconds before the break is enough for me, but the sound alert is too much and I can't turn it off.

hovancik:
Hi there and thanks for the feedback. 

Yeah, the sound scheme is sound at the end of break. 

It's not possible to disable notification sound from app, AFAIK. To get rid of sound you can disable whole notifications at 3th settings page ;/ 

There might be some setting in MacOS to turn off notifications sound, don't know. (There is such a option in WIndows)

asterite:
@hovancik Thank you! Don't worry, I will send a PR for this so you have an initial codebase :-)

hovancik:
closing ;] thanks for PR

PR #329, opened 2018-11-01
Add "silent notifications" setting

asterite:
Issue: #327

### Requirements

- [ ]  during development, `node` version specified in `package.json` was used.

Explanations to why some things are unchecked:
- I opened an issue but forgot to discuss the implementation. I describe the implementation below. If you don't agree with it you can close this PR, or use it as a basis for another implementation.
- I couldn't find an easy way to install that specific node version (I actually have an older `node` version and then I upgraded to a newer `node` version, and in both cases it worked fine) but the changes are relatively small that it shouldn't break depending on the version
- I didn't add tests to the code, but it didn't have tests either
- `npm run lint` gives errors, but it gave errors before (I can fix this in a next PR if you want)

### Description of the Change

This PR adds a new setting to the 3rd settings page: "silent notifications". It's unchecked by default, so the default behavior is the same as before this PR. If checked, the OS notifications that pop-up are now silent.

I thought about doing this only when the "silent" theme is selected. However, making it in this way is more flexible because you can have a silent theme but sound-y notifications, or the other way around.

Additionally, ~in a separate commit~, I extracted the "show notification" logic to a separate function because some duplicated code started to show.

I added default English texts for the language translations expect for Spanish because I'm a native Spanish speaker (from Argentina). In a next PR I can send the missing Spanish translations if you want.

Oh, I unchecked one of the TODO checkboxes, but I'm not sure that's right. I can revert that change if needed.

### Verification Process

To verify that it worked I:
- left the new checkbox unchecked and confirmed that a sound appears when showing any of the notifications (break, microbreak and resume)
- left the new checkbox checked and confirmed that a sound **doesn't** appear when showing any of the notifications (break, microbreak and resume)

I assume the changes didn't introduce any regressions because they only modify that part of the app's logic. I also made sure to update all code that's related to the `showNotification` logic.

### Other information

Thank you for the app, it's great! The least I can do is try to improve it :-)


asterite:
I will rebase in a few hours

app/utils/defaultSettings.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
   posLatitude: 0.0,
   posLongitude: 0.0,
   useMonochromeTrayIcon: true,
+  silentNotifications: true,
```
hovancik: by default should be false, so we do not change the existing behavior

hovancik:
Many thanks for PR ;] I will check the code. 

For different versions of node, I am using nvm https://github.com/creationix/nvm. 

Feel free to make Spanish translations better ;]

I have question: 

What about new version notification? Do you think we should also use this settings for it or not?

app/utils/defaultSettings.js
asterite: Fixed!

asterite:
Hi! I just rebased the code against `master`.

About the new version notification, I'd use the same preference. It's either all notifications on for the app or all off. That is, unless you want to have specific settings for each notification, but I think a single one is simpler. In general I think you either like the notification sound or you don't, globally for the app.

I'll make that change right away so it's consistent with everything else, but let me know if you'd like it some other way.

hovancik:
I would do the same, but didn't want to influence you ;]

asterite:
Done! Let me know what you think.

And thank you for the recommendation of `nvm`, it worked out of the box. Maybe you can suggest to use it in the issue template :-)

PR #333, opened 2018-11-06
Fix 'pause' Spanish translation

asterite:
Hi!

Sorry I didn't fill all the required issue details.

The current Spanish translation for "pause" is "Descanso" which in English is "break" or "rest" and that's used for the "Pause -> for an hour" menu and so on, so the translation isn't quite correct. The correct translation is "Pausar", which means "To pause".

This translation existed before I added the missing ones. I noticed it when using the app in Spanish :-)


hovancik:
do you wanna add yourself to contributors in readme? ;]

hovancik:
Ah, cool ;] I should have checked :D

PR #452, opened 2019-08-19
Tweak/keep focus on active window

mfyz:
### Requirements

No, this is not an issue, a minor tweak on the user experience.

It is only failing 2 test case where it is also failing identically on master branch. The minor tweak I made is not affecting any existing functionality and it doesn't fail any tests different than the master branch. 

Here is the failing test scenario:

```
1) UntilMorning
       Sunrise Settings
         loadMorningTime() returns morning time when sunrise is an input:

      AssertionError: expected 8 to equal 9
      + expected - actual

      -8
      +9

      at Context.<anonymous> (test/untilMorning.js:106:31)
      at processImmediate (internal/timers.js:439:21)
      at process.topLevelDomainCallback (domain.js:126:23)
```

### Description of the Change

When showing break windows, if I'm typing/coding, my input is broken right away. I may be finishing up a sentence or a piece that my keystrokes shouldn't be interrupted. Showing windows with "showInactive" achieves this.

### Verification Process

This is a minor change that continues to utilize electron native capabilties. There is also no other functionality requires the break windows to be active (they are already "on top" just not taking the focus over from the previously active window).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - German translations updated
 - no notification is shown after system resume/unlock
 - no notification is shown after manual resume of pause from tray menu
+- when showing windows, keeping the focus on active window, so typing behavior won't stop when displaying break windows
 
 ### Added
 - pause breaks when screen is locked (Windows, macOS)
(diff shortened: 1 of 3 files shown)


hovancik:
Hi, I'll try testing on Linux but macOS looks good. 

Tests are passing on CI so it's OK. It's probably some local issue and everything is also fine on my end.

PR #454, opened 2019-08-19
fitscreen advanced setting added

mfyz:
### Requirements

This is a new change, not related with an issue.

### Description of the Change

This is a minor addition to advanced settings where the default non-fullscreen break windows is displayed 800x600 sized windows. Having full screen enabled causes macOS to treat each full screen window as a brand new "space" in the expose/mission control. This is too much change and feels heavier when macos does animate this behavior heavily. Keeping windows not full screen is also not obstructive enough to make me stop working, sometimes writing and seeing cursor below the break window can be ignored easily.

I've added a flag in settings (false by default, so no change of behavior by default). The flag sets break window size and position to cover all "work area" size (fit screen except menubar on macOS, full screen on other OSes).

I also moved the common block that determines display's bound size that is used in displaysX and displaysY to a new method displaysBound method which should reduce deduped code.

### Verification Process

Tested in different screen sizes, multiple monitors/external. Also this feature's setting is disabled by default, so there won't be any change in standard behavior of stretchly app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - pause breaks when screen is locked (Windows, macOS)
+- fitscreen advanced setting is added
 
 ## [0.20.1] - 2019-07-14
 ### Added
--- a/README.md
+++ b/README.md
@@ -100,6 +100,9 @@ In the settings file, set `notifyNewVersion: false,` to disable new version noti
 #### Play sound at the start of the micro/break
 In the settings file, set `microbreakStartSoundPlaying: true,` to start a microbreak with a sound. Same for `breakStartSoundPlaying`.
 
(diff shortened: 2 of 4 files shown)


hovancik:
which OS did you check? Doesn't seems to work on Debian XFCE for me (it is only on one screen and covering around bottom 60 percent)

mfyz:
it was macos. I can try on a virtual machine ubuntu when I get a chance. this feature is also primarily focusing on mac users since full screen is already the proper feature for this - but macos window manager’s full screen behavior makes this annoying, especially this frequent app (seeing a full screen disruptive and long animation in every 10 minutes is not ideal)

hovancik:
It would be nice if it would work on Linux because fullscreen does not work there. Also if it's electron bug, would be good to report it.

hovancik:
closing as with redesign similar fix has been applied

Issue #420, opened 2019-06-28
Minor UX Issues


morinted:
### Prerequisites

- [x] I'm using latest version: v0.19.1
- [x] I've checked [Known issues](https://github.com/hovancik/stretchly#known-issues)

### Description

The UI has very thin font and the color schemes are low contrast, leading to some usability issues. This is exacerbated by Mac display scaling on an external 4K monitor.

Additionally, in the settings, the buttons don't really look like buttons, and the radio buttons for the audio and color schemes are not following conventions, making them hard to use.

I get the sense that a lot of the design decisions are aimed towards creating a Zen aesthetic or personality, but I fear that the user experience isn't the best.

Operating system: Retina macOS

### Proposed adjustments

- Use an easier to read body font—either a different font or a heavier weight.
- Adjust the controls on the schemes settings page so that the radio buttons look more like radio buttons (highlighting the selected option with a glow, add a smaller box to indicate selection, hover states when your mouse is over them)
- Make a more consistent button styling (+, -, reset to defaults, skip this break) where it's clear to the user that these are buttons (hopefully without sacrificing aesthetic.)
- Increase font color contrast. The `coffee kisses` theme should probably use a black font. If this is implemented, adding a lighter theme would also be possible.

I'd be willing to assist in implementing some of these changes as I really like the work that's gone into this application.

hovancik:
With new redesigned version coming soon, I am closing this :)

PR #488, opened 2019-10-23
Make buttons more obvious

richyliu:
Issue: #420 

### Requirements

### Description of the Change
I added a inset shadow and a drop shadow on hover to the "+" and "-" buttons on the settings page. I also changed the checkboxes to this style as well. This should make it more obvious to the user that they are buttons.

Screenshot from 2019-10-22 20-14-27

### Verification Process
I tried clicking on the buttons to make sure the click listeners still work. I tried checking the checkboxes to make sure they still save the checked state.

### Other information
I noticed some other checkboxes in `app/contributor-settings.html`, but I didn't change them because I didn't know how to access them. I didn't change the "skip this break" button on the break timer because it didn't look too well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Improved grammar of microbreak ideas
 - display checkbox list vertically in settings screen 3
 - Dutch translations updated
+- drop shadow on hover for time changing buttons
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


hovancik:
Hi and thanks! 

I like the checkboxes but not the +- buttons. Not sure what to do with those ;/ Do you have a different idea about +- maybe? 

Let's see what other people thing as well. 

(Anyway good news: I am working with designer to rework whole app from ground up! )

hovancik:
Closing this as redesign is underway in #537

PR #596, opened 2020-05-28
Fix Linux fullscreen

### LINKED ISSUE #186: Impossible to read texts with "breaks are fullscreen" on Elementary OS 4.1 (Loki)
Hello,

I'm sorry if my bug is poor of details. I've just installed stretchly 0.14 in Elementary os 0.4.1, 64 bit.

As you can see opening the attached image, messages and images of the app are almost impossible to read. It seems that the image is unaligned ... with the window itself.

In settings "breaks are fullscreen" is checked (everything works well if it's unchecked).

Does it depends on my system or is it really a bug?

screenshot from 2018-01-13 18 23 32

hovancik:
closes #186


app/main.js
```diff
@@ -332,16 +376,28 @@ function startMicrobreak () {
...
       windowOptions.x = displaysX(displayIdx)
       windowOptions.y = displaysY(displayIdx)
     }
 
+    console.log('monitors id', displayIdx)
+    console.log('monitors windowOptions', windowOptions)
     let microbreakWinLocal = new BrowserWindow(windowOptions)
     // microbreakWinLocal.webContents.openDevTools()
+    console.log('microbreakWinLocal.Bounds after creation', microbreakWinLocal.getBounds())
+    microbreakWinLocal.setSize(windowOptions.width, windowOptions.height)
```
hovancik, line 394: @dbankmann do we need this? it causes error in regular sized window as `windowOptions.height` does not exist

we are settings it on line 379 so I would say this is duplicated

dbankmann: Does this work?
```suggestion
    microbreakWinLocal.setSize(widht = windowOptions.width, height = windowOptions.height)
```
I removed the call via keywords in the last commit, as it produced linting errors.

This line is duplicate indeed, but if I remember correctly this was the essential part of the multi-head fix. Somehow the constructor `BrowserWindow` doesn't get the settings right. I'm not at my multi-head machine atm, so cannot test.

hovancik: You mean multiple displays? Will try to test today. Electron sucks in this that some parameters seems to work only when called via method or vice-versa ;/ 

This is API, should be as you did in the end: https://www.electronjs.org/docs/api/browser-window#winsetsizewidth-height-animate 

```
win.setSize(width, height[, animate])

```

Anyway, the problem there is that windows.height does not exist as I am not in multiple display situation (line 484)

hovancik: I've added default width, height, so we don't get error, Kept `setSize` just in case

dbankmann:
Do you mind adding myself to the contributors list? As I didn't start the PR myself, its not recognised by github...

hovancik:
Yeah, sure :) go ahead

PR #597, opened 2020-05-28
Hardcore mode

balazsnasz:
### Requirements

### Description of the Change

As we discussed previously in email, I implemented a Hardcore mode feature.
Hardcore Mode disables Pause breaks, Reset breaks, Postpone/Skip breaks, Quit app features.

As we also discussed, I will create two separate PRs of this. First, this one is without GUI and can be set in the config.json. And the second will have a GUI.

Please test it before merging.
I tried to test it for a while, and every possible settings (strict mode on/off, postponement on/off, hardcore mode on/off, also test ctrl + x when hardcore mode is on).


app/main.js
```diff
@@ -731,16 +735,21 @@ function getTrayMenu () {
...
     label: i18next.t('main.quitStretchly'),
+    visible: !settings.get('hardcoreMode'),
     role: 'quit',
     click: function () {
       app.quit()
     }
   })
 
-  return Menu.buildFromTemplate(trayMenu)
+  // Hide the invisible items. Electron has a bug which ignores visible: false parameter for separator menuitems. Hide them here.
```
hovancik, line 749: can you point me to bug? I would like to subscribe to it

app/utils/utils.js
```diff
@@ -54,13 +54,13 @@ const formatTimeIn = function (milliseconds, i18next = require('i18next')) {
 }
 
 // does not consider `postponesLimit`
-function canPostpone (postpone, passedPercent, postponePercent) {
-  return postpone && passedPercent <= postponePercent
+function canPostpone (postpone, passedPercent, postponePercent, hardcoreMode) {
```
hovancik, line 57: so in HC mode, I am not able to postpone or skip?  

`tests/utils.js` should be updated with new variable and test cases.

app/main.js
```diff
@@ -686,6 +688,7 @@ function getTrayMenu () {
   } else if (!doNotDisturb) {
     trayMenu.push({
       label: i18next.t('main.pause'),
+      visible: !settings.get('hardcoreMode'),
```
hovancik, line 691: I think we should hide this options only when in break? 

Usecase for hardcore mode for me seems to be "make it even harder not to finish break early". I think I would still want to pause breaks when I to present something, or reset.

app/main.js
```diff
@@ -731,16 +735,21 @@ function getTrayMenu () {
       createPreferencesWindow()
     }
   }, {
-    type: 'separator'
+    type: 'separator',
+    visible: !settings.get('hardcoreMode')
```
hovancik, line 739: Not sure about hiding Quit button. That's something that should always be there, it's basic functionality.

balazsnasz: What do you think about adding a new config item then?
Breaking hardcore mode into separate config items such as: disable reset breaks, disable skip & postpone breaks (it doesn't make that sense, because we already have strict modes), disable quit app features, disable pause breaks.

In my opinion, real hardcore mode makes it impossible to avoid breaks.
I really need this feature in the implemented way because I always find ways to avoid breaks. 😁

Implementing separate config items seems a reasonable way to handle this issue because then everyone can decide which features they want to hide/disable.

app/main.js
balazsnasz: In my opinion, real hardcore mode makes it impossible to avoid breaks.
I use the full screen layout, so hiding the pause/reset breaks menu items doesn't make sense for me.
I can pause/reset anytime then, so it's not a real hardcore mode IMO.

app/main.js
balazsnasz: https://github.com/electron/electron/issues/6906
https://github.com/electron/electron/issues/3494

app/main.js
hovancik: I am pretty sure I don't wanna multiple settings, it would make every new feature more complicated. I've made that mistake few times already...

I think we need to move a step back and spend some time thinking what does "hardcore mode" means...

balazsnasz: For me Hardcore mode means the following: It doesn't allow you to avoid mini/long breaks in any way. It not just disables skipping breaks, it also hides the reset & pause break features, and making it almost impossible to quit/skip breaks and the app itself.

hovancik: So I was thinking last few days and here's what I am leaning towards now: 
- having two settings for one thing is bad idea, so I wouldn't let this mode change postpone/end breaks functionality. User can go hardcore by disabling both of them right now
- selectively picking what to hide in tray seems wrong to me as well, different user will have different opinions what hardcore means and they will be asking me to change the behavior 

So what about adding simple `tray`, that would hide/show whole tray? In this way there is no discussion what hardcore means, we simply have option to hide whole tray. 

Considering that tray is then gone, user needs to go and edit settings file. Pretty hardcore for me :robot:

balazsnasz: Hiding the tray icon doesn't seem like a good solution for me because that's the one and only thing that displays how long time if left until a next break.

However if we can meet in the middle line that would be a good compromise:
Show the tray icon but disable the tray menu completely. So the display of remaining time until next mini/long breaks are shown in the tooltip, but you can't change anything from the menu/quit the app.

What do you think of this?

hovancik: Let me think a bit more :D Maybe I think of something but probably that's it.

balazsnasz: Did you find a better solution for this?

hovancik: Not really :) Let's add `trayMenu` `true/false` that will hide tray menu. But let's keep the status messages as not every platform is bug-free with tray tooltip.

balazsnasz: What do you mean by status messages?
The one which is in the top of the tray menu?

So, should I just hide the menu items in the tray menu, except the top ones which shows remaining time?

How should we turn on/off hardcore mode via the UI? When I originally developed this feature I also implemented a checkbox element for HC mode in Contributor prefs window.
If the Preferences tray menu item is also hidden, then you won't be able to turn off hardcore mode from the UI. Any ideas how to solve this?

hovancik: Yeah, I mean those messages at the top of the menu. 

Well, I would have it only in the file. Makes sense for me to complicate setting/unsetting it :)

PR #600, opened 2020-05-29
Ability to change keyboard shortcuts via config.json

### LINKED ISSUE #497: Stretchly Intercepting Ctrl + X on Windows 10 on v0.21.0
### Prerequisites

- [✓] I'm using latest version: v0.21.0 
- [✓] I've checked [Known issues](https://github.com/hovancik/stretchly#known-issues)

### Description

For some time I've had an intermittent issue where the ctrl + x hotkey (commonly "cut to clipboard") stopped working in random applications.  Finally I got tools installed to track the problem when next it happens.  Today I once again ran into the issue, started up Spy++, and used its message monitoring capabilities to track the WM_HOTKEY message all the way to Stretchly:

image

Sure enough, when I closed Stretchly the hotkey interception went away.

Now, while typing all this up I was going through the issue checklist and saw that version 0.21.0 was supposed to have `Fixed: Ctrl+X global shortcut not being released after Reset breaks and Skip to`, but I think a logic branch was missed.

After playing with it a bit, it seems that if "Reset breaks" is used while a (regular) break is currently on the screen, it causes the issue.  Skipping during a microbreak seems to do nothing.  

When manually skipping to the next full-sized break and hitting Reset Breaks, I sometimes get the following error message as soon as I next hit ctrl + x:

image

In which case ctrl + x is not bound, but if the error message doesn't show, ctrl + x is bound until Stretchly is restarted.  I suspect the behavior is slightly different when a break happens naturally rather than manually invoked, but did not confirm.  

Operating system: Windows 10 Pro 1809

### Steps to Reproduce

1. Skip to the next break (not microbreak)
2. Reset breaks using the taskbar icon context menu while the window is on-screen
3. Use the ctrl + x hotkey.  If an error message does not show, ctrl + x is now bound permanently by Stretchly until restart.  

**Expected behavior:** ctrl + x is never bound except when the break screen is showing

**Actual behavior:** ctrl + x is intercepted and unusable by other applications

**Reproduces how often:** possibly every time a (full-sized) break is skipped, modulo whatever controls that error message showing up

### LINKED ISSUE #594: Windows 10 Shortcuts
Hello! 

Thanks for the app ! I'm using it daily and I appreciate the breaks, the breeze feeling and the ease of use and set up.

I'm using Ctrl+X to cut a lot when I'm typing texts on editors. It has happens sometimes just when the popup comes up and well it postponed the break.

I'd like to have a way to disable the shortcut or edit it to be in a less used/accessible shortcut.

Thanks for your attention and have a nice day !

balazsnasz:
Added ability to change the keyboard shortcuts for Skip/Postpone breaks via config.json

Issue: closes #594
closes #497  

### Requirements

- [X]  issue was opened to discuss proposed changes before starting implementation.

### Description of the Change
I added the ability to change the keyboard shortcuts for Skip/Postpone breaks via config.json.
The default is still Cmd/Ctrl + X, but you can change to Alt, Shift, Cmd/Ctrl, and any a-z A-Z 0-9 keyboard shortcuts.

GUI is coming soon in a different PR.
Be careful when you accept this PR because it will have a conflict with the Hardcore Mode PR (#597 & #599). I changed the on('progress') parts for both of those - I added parameters to the function, so be careful to keep the order in all of the occurrences.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Added
 - Added Hungarian translation
+- Added ability to change the keyboard shortcuts for Skip/Postpone breaks via config.json
 
 ### Changed
 - update Chinese (Traditional) translation (#590)
--- a/README.md
+++ b/README.md
@@ -106,6 +106,15 @@ In the settings file, set `naturalBreaksInactivityResetTime` to your preferred v
 #### Volume for break sounds
 In the settings file, set `volume` to your preferred value. Default value is `1`, which is 100% volume. Set it, for example, to `0.61` for 61% volume.
 
+#### Postpone/Finish Break Shortcut
+In the settings file, set `endBreakModifier` and `endBreakKey` to your preferred values. Shortcut is then applied as `endBreakModifier+endBreakKey`. For available values for key and modifier check [Electron's documentation](https://www.electronjs.org/docs/api/accelerator) as we do not validate your values.
+
+Default values, that will result in `Cmd/Ctrl + X` as a shortcut:
+```
+endBreakModifier: 'CmdOrCtrl',
+endBreakKey: 'X',
+```
+
 ## Install [![Github All Releases](https://img.shields.io/github/downloads/hovancik/stretchly/total.svg)](https://github.com/hovancik/stretchly/releases/latest)[![Packaging status](https://repology.org/badge/tiny-repos/stretchly.svg)](https://repology.org/project/stretchly/versions)
 
 Latest installers for macOS, Windows, Linux and FreeBSD can be found [here](https://github.com/hovancik/stretchly/releases).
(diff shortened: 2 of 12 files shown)


hovancik:
@balazsnasz I've made some changes to your code: 
- more generic name for shortcut, as there will be more of them in future
- no validation and let Electron take care of that, so we have more options for values
- added test
- simplify some code

I think I can merge if it looks good to you.

balazsnasz:
@hovancik I checked it, and it seems fine. You can merge it.

hovancik:
Now I am thinking, maybe I should just have `shortcut`, without separating it to `key` and `modifier`... 

Not sure if I'll ever need them separated, and even then it's easy to split by `+`

balazsnasz:
Probably we won't ever need them separately. But when you think about creating the GUI for this feature, it's a little easier to update the separate setting values than updating them together.

I thought I would build the GUI like this:
[Modifier Dropdown] + [Key Dropdown]

Setting (e.g. saving) Dropdowns is already implemented in the Preferences & Contributor Preferences windows, so it would easy to add such interface elements.
However if we'd have one config element (like shortcut or so), we would handle the saving/loading the element's values as an exception.

Not that big issue though, so if you think we should have only one config item, then it's fine.

hovancik:
I've check for example Atom, and they have no UI to change shortcuts, only ability to edit config file. I went with more simple solution over premature optimization. There will need to be research about how shortcuts are managed in UI elsewhere, if there is need for it.

Issue #503, opened 2019-12-09
Disabling microbreaks does not stop the next one


Quitch:
### Prerequisites

- [X ] I'm using latest version: v0.21.0
- [X ] I've checked [Known issues](https://github.com/hovancik/stretchly#known-issues)

### Description

When disabling microbreaks in settings, if a microbreak is the next scheduled break type, the schedule is not updated until that microbreak has occurred.

Operating system: Windows 10 Pro x64 1909

### Steps to Reproduce

1. Ensure a microbreak is scheduled to occur next
2. Disable microbreaks

**Expected behavior:** The next break will be a break

**Actual behavior:** The scheduled microbreak will occur

**Reproduces how often:** 100%

### Additional Information

Resetting breaks after making the change correctly updates the schedule

hovancik:
Hi, you are right. It is like that with more settings, ie when you change time between breaks it is changed only after next break. My thinking was that I don't wanna reset breaks everytime someone changes the settings, but what you say makes sense, it should be changes right away.

Quitch:
Ideal behaviour would be that the break timings are unaffected and microbreaks simply removed from the schedule, but an automatic reset would work too.

hovancik:
closing in favor of #1248

PR #601, opened 2020-05-29
Enabling/disabling Mini/Long Breaks resets the breaks

balazsnasz:
Issue: #503

### Requirements

- [X]  issue was opened to discuss proposed changes before starting implementation.

### Description of the Change
Enabling/disabling Mini/Long Breaks resets the breaks (#503)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - update Chinese (Traditional) translation (#590)
 
+### Fixed
+- Enabling/disabling Mini/Long Breaks resets the breaks (#503)
+
 ## [0.99.4] - 2020-05-15
 ### Fixed
 - checkbox across macOS, Windows, Linux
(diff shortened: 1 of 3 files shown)


hovancik:
Let me check because there are few possible places to do this ;]

hovancik:
I think I would like to have more complex solution. First thoughts: 
- it should be in `main.js` `ipcMain.on('save-setting', function (event, key, value) {` where I do specific stuff for specific settings, so it's kept in same place as other special things
- I think we should probably have special class in utils to handle changes, so that we can test it, depends how much code it ends up having (and if there's even a way how we could test it)

Some quick pseudo (could be some methods that we call in `save-settings`)

```
- when enable/disable break is changed
  - check if next scheduled is enabled  
    and remove from scheduler, schedule next enabled if needed
- when enable/disable notification is changed
   - check what's next and schedule next break with/without notification
```
with scheduling new break I would like to consider how much time has already passed, so if next scheduled break is Mini in 5 minutes, and we are disabling notification, it will be in 5 mins but without notification   

What do you think?

balazsnasz:
Are you sure you want to make this so complex?
Change such settings are not a too regular thing.

hovancik:
Well, I am not sure :D But I think it's expected. 

Let's say when I am 1 minute to Mini and I disable Long, it's unexpected that breaks are reset.

PR #652, opened 2020-08-05
Handle windows focus assist

### LINKED ISSUE #645: Stretchly doesn't respect Focus assist setting on Windows 10 (aka how to enable do-not-disturb on Windows)
### Prerequisites

- [x] I'm using latest version: v1.0.0
- [x] I've checked [Open issues](https://github.com/hovancik/stretchly/issues)

### Description

Currently, no matter what option is set for _Focus assist_, Stretchly will still show its reminders.  
So it seems there is no way to temporarily disable the overlays during a presentation, video call, or similar.

Maybe this is only a documentation issue, but I could not find any instructions on how to enable do-not-disturb on Windows 10 so Stretchly recognizes it.

Operating system: Windows 10 v1909 build 18363.959

### Steps to Reproduce

1. Set the shortest possible schedule for mini breaks
2. Enable Focus assist:  
image
3. Wait for scheduled break

**Expected behavior:**
Stretchly doesn't interrupt when one of the focus modes is selected.

**Actual behavior:**
Stretchly doesn't care 😄 

**Reproduces how often:**
100%

Aziks0:
Issue: closes #645 

### Requirements

### Description of the Change
Focus Assist mode is available since the Windows 10 version 1803.

image

For W10 version above 1803:
- When the Windows Focus Assist mode switches to "Priority only" or "Alarms only", Stretchly breaks are paused
- When the Focus Assist mode switches to "Off", breaks are resumed

For W10 version older than 1803/Linux/MacOS:
- Stretchly works exactly like before

### Verification Process
I've tried on Windows v1903 and it works as expected:
- When Focus Assist mode is on "Priority only" or "Alarms only", breaks are paused
- When Focus Assist mode is on "Off", breaks are resumed

I've tried on Windows v1703 and it works like before:
- When DND (Quiet Hours) is on, breaks are paused.
- When DND (Quiet Hours) is off, breaks are resumed.

On Linux and MacOS, it will normally work like before.

### Other information
[Known Issue #1](https://github.com/hovancik/stretchly#known-issues) might be resolved if users upgrade to a Windows 10 version above 1803.

### DIFF
--- a/app/utils/dndManager.js
+++ b/app/utils/dndManager.js
@@ -26,7 +26,12 @@ class DndManager extends EventEmitter {
 
   get _doNotDisturb () {
     if (this.monitorDnd) {
-      return require('@meetfranz/electron-notification-state').getDoNotDisturb()
+      let focusAssist
+      try {
+        focusAssist = require('windows-focus-assist').getFocusAssist().value
+      } catch (e) { focusAssist = -1 } // getFocusAssist() throw an error if OS isn't windows
+
+      return require('@meetfranz/electron-notification-state').getDoNotDisturb() || (focusAssist !== -1 && focusAssist !== 0)
     } else {
       return false
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## Added
+- breaks are paused if the Windows 10 Focus Assist mode is enabled
+
 ### Changed
 - updated Polish translation
 - updated Chinese translations
--- a/README.md
+++ b/README.md
@@ -236,6 +236,7 @@ You can copy debug information to the clipboard.
 - Felix W. Dekker, [@FWDekker](https://github.com/FWDekker)
 - Balazs Nasz, [@balazsnasz](https://github.com/balazsnasz)
 - Daniel Bankmann, [@dbankmann](https://github.com/dbankmann)
+- Aziks, [@Aziks0](https://github.com/Aziks0)
 
 Also see Github's list of [contributors](https://github.com/hovancik/stretchly/graphs/contributors).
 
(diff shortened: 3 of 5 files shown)


hovancik:
Could we simplify this? I don't think we need to duplicate the code, we should have one timer only and in `_doNotDisturb` simply do `||` for the `electron-notification-state` value and `windows-focus-assist`. If one of those says we are in DND, we are then in DND.  

It will be easier to read and we don't need to update tests.  

What do you think?

Aziks0:
Yeah it's a good idea to only use `_doNotDisturb`, I will simply it.

Though, I think I can't simply use `||` to return the value of `_doNotDisturb` because `getFocusAssist()` can return 4 things:

1. `{ value: -1, name: "FAILED" }`
2. `{ value: 0, name: "OFF" }`
3. `{ value: 1, name: "PRIORITY_ONLY" }`
4. `{ value: 2, name: "ALARMS_ONLY" }`

and when Focus Assist isn't available, the object `{ value: -1, name: "FAILED" }` is always returned so `return require('@meetfranz/electron-notification-state').getDoNotDisturb() || require('windows-focus-assist').getFocusAssist().value !== 0` will always return `true` on Windows version older than 1803, even if DND (Quiet Hours) is off.

app/utils/dndManager.js
```diff
@@ -5,7 +5,7 @@ class DndManager extends EventEmitter {
     super()
     this.settings = settings
     this.monitorDnd = settings.get('monitorDnd')
-    this.timer = null
+    this.dndTimer = null
```
hovancik, line 8: One more thing: let's rename this back to `timer` :) not really important but... you know... :) 

could you also squash all your commits to one? if not, I can also do that when merging :)

hovancik:
Hey @Aziks0 ,

since this library was added I am not able to run `npm i` on Windows: 

image

Have you encountered this? How to fix it?

Aziks0:
No issue on my PC, no issue on my VM.
Although, according to this [stackoverflow question](https://stackoverflow.com/questions/3898287/c-include-atlbase-h-is-not-found), `altbase.h` comes with a C++ Visual Studio component which is installed on both my PC and my VM.

hovancik:
For future reference: 

I think I needed to go to "Add remove programs" , find "VS Build Tools 2017" , then "Modify" and select "ALT something something"

PR #661, opened 2020-08-18
Added ability to set the opacity of breaks windows

### LINKED ISSUE #657: Add an option to change the opacity of the breaks windows
### Prerequisites

- [x] I'm using latest version
- [x] I've checked [Open issues](https://github.com/hovancik/stretchly/issues) (and [Closed issues](https://github.com/hovancik/stretchly/issues?q=is%3Aissue+is%3Aclosed))

### Description

Sometimes a break start while I'm writing, and since the window is not transparent, I cannot finish writing down my sentence/line of code before I take my break because I cannot see what I'm writing. In this case, I don't really want to postpone my break because I only need a few seconds to finish writing.

So it could be really nice to have an option to add transparency, with the possibility to change the opacity percentage (maybe a Contributor Preferences only ?).

I already tried changing the opacity and it looks like this with an opacity at 0.5 (with Green Clouds theme):
Opacity with Windows background

However, depending on the background, it may be hard to read due to the white font color:
Opacity with black and white background

Aziks0:
Issue: closes #657 

### Requirements

### Description of the Change
I added the ability to set the opacity of the breaks windows.

A new option is available in the `Theme` tab:
image

The value can be set from 0.1 to 1, with a step of 0.1, and is set to 1 by default.

### Verification Process

I tested it on Windows 10 v1903 and on Ubuntu 16.04, it works on both OS.

# Issue

image
If the value is a float (so when opacity is set to `0.x`), the value displayed will be `0`. This is because the value displayed is converted to an integer.
###### see `app/preferences.js`, line 190 & line 193: ```output.innerHTML = i18next.t(`utils.${unit}`, { count: parseInt(range.value) })```

Do you have any idea how to fix this ?

I though about 2 things:
- another `querySelector` in `app/preferences.js` for the opacity value only
- use percent (so integer between 10 and 100) instead of float, but with this approach, every time you call `settings.get('opacity')` you need to divide the returned value by 100 to get a float back.


app/preferences.html
```diff
@@ -256,6 +256,21 @@
...
+            <option value="0.2"></option>
+            <option value="0.3"></option>
+            <option value="0.4"></option>
+            <option value="0.5"></option>
+            <option value="0.6"></option>
+            <option value="0.7"></option>
+            <option value="0.8"></option>
+            <option value="0.9"></option>
+          </datalist>
+          <output data-unit="none"></output>
```
Aziks0, line 272: I didn't specify any unit because I don't know which one I could use, is there even one ?

hovancik:
Hi @Aziks0 , thanks for the  PR. I will write to UX guy and ask about the best place/ux for opacity. 

I'll also ask him about defaults, whether we should go for .9 or something :) Let me know what you think as well.

After I'll answer the other questions

hovancik:
Hey @Aziks0 , I got answer from designer :) 

1. default should be solid (so what you did) as opacity is bad for accessibility
2. he recommends doing similar to what Apple does: have checkbox for transparency (which changes between solid and transparent background)   

What do you think? I think I agree :) 

We would then have: 
- in Preferences checkbox "Enable transparency for breaks / Enable transparent breaks / Enable transparent background" (or something similar) that would change `transparency` in config (default: `false`)
- in Contributor Preferences we would have "Opacity: if transparency is enabled in main Preferences, you can specify percentage for opacity" that would change `opacity` in config file, (default: `0.9`)

For opacity, we should store 0-1, but to user show 0-100 percent. (You can see how I did that with volume in Contributor preferences. You can easily get there by setting `isContributor: true` in `main.js`)

Question: do we need to change `mainColor: '#FF478484',` in config?

Aziks0:
Yeah it was my initial thought, so I completely agree. Do you want me to put the Opacity option at a specific place in Contributor Preferences ?

> Question: do we need to change mainColor: '#FF478484', in config?

No it's not needed, and I guess having `mainColor: '#FF478484'` will break the default settings for the Apparence Theme. I didn't thought about this, I will revert it back.

hovancik:
:+1: glad to hear it

just put it where you think it fits the best :)

Aziks0:
@hovancik Ok, I did something wrong with this PR but I don't know exactly what I did😅

hovancik:
Hey no worries, it was probably me merging another PR to master :)

Aziks0:
There is now a checkbox in the `Theme` tab, under `Appearance`:
electron_xZKjXwErhf
It works as you asked, and it's off by default.

There is now a slider in `Contributor Preferences`, for changing the opacity value:
electron_nLFcpq15hy
It also works as you asked, and it's set to 0.9 by default.

I might change `Theme opacity:` to `Theme transparency:` in `Contributor Preferences` for consistency, and I don't know if the info is clear enough, what do you think ?

app/contributor-preferences.html
```diff
@@ -258,6 +258,30 @@
...
       </div>
+      <div>
+        <span data-i18next="contributorPreferences.opacityTheme"></span>
+      </div>
+      <div>
+        <p data-i18next="contributorPreferences.opacityInfo"></p>
+      </div>
+      <div>
+        <label data-i18next="contributorPreferences.opacity" data-same-width-label for="opacity"></label>
+        <input type="range" min="10" max="100" step="10" list="tickmarksOpacity" name="opacity" data-divisor="0.01" id="opacity">
```
hovancik, line 269: Could we make a step 5? 10 seems to small... And also make min 5 as well? Pretty much I think I would make it the same as Volume so that it's consistent.

hovancik (changes requested):

hovancik:
> I might change Theme opacity: to Theme transparency: in Contributor Preferences for consistency, and I don't know if the info is clear enough, what do you think ?

Yeah, I am not sure. Are you a native English speaker? I guess they should be able to say

Aziks0:
> > I might change Theme opacity: to Theme transparency: in Contributor Preferences for consistency, and I don't know if the info is clear enough, what do you think ?
> 
> Yeah, I am not sure. Are you a native English speaker? I guess they should be able to say

No I'm french, I feel like the sentence is a bit weird but my main concern is about the wording:
- `opacity` or `transparency` ?
- `break window` or `theme` ?

I don't know which words are the most explicit about what the option does.

hovancik:
Let me ask around and get back to you

hovancik:
So, how about this? 
- `Enable transparency`
- `Theme transparency: How solid should the break window be?`

Aziks0:
Yeah, `solid` explains well what the option modifies. I think there will be some translation issues tho, at least in French because there isn't any good translation for the word or the sentence. But I'll figure it out.

hovancik:
Hmm right, Slovak would be also difficult.

Aziks0:
So we can keep the idea of talking about `break window` instead of `theme` and go back to something like `you can specify the opacity percentage of the break window`. It's a little less clear than what you proposed, but I guess it's clear enough and there will be less (if any) translation issues.
I think we are overstating the difficulty of the sentence, but I don't know 😅

hovancik:
Let's use what I suggested? After thinking about it, it's fine that each language will use different wording. If it makes sense in English, it doesn't mean that other language should be 100% copy.

Aziks0:
Yeah, true.
I've already made the changes in my last commit so everything should be good.

app/locales/en.json
```diff
@@ -130,6 +131,9 @@
     "pauseUntilMorningInfo": "Specify till what hour 'Pause until morning' should pause.",
     "welcomeWindow": "Welcome window:",
     "showWelcomeWindow": "Show Welcome window on next start",
+    "opacityTheme": "Theme transparency:",
+    "opacityInfo": "If transparency is enabled in Preferences, how solid should the break window be ?",
```
hovancik, line 135: Only small change here: `How solid should the break window be?` According to native speakers, the `If ...` is not needed as it's pretty clear what we are talking about. 

Do you know how to squash commits to one? I think I can merge after :)

hovancik (changes requested):

app/locales/en.json
Aziks0: Do you want me to also remove the `if ...` in the French translation ?

hovancik: Well I guess depends on the translator :) What do you think is more clear there...

Aziks0: I prefer with the `if`, so I'm gonna leave it like that.

Aziks0:
I can't rewrite my git history, can you use github to squash everything ?

hovancik (approved):

hovancik:
Thanks!

Issue #667, opened 2020-08-27
FR: Daily limit clock ?


mwoz123:
### Prerequisites

- [x ] I'm using latest version: v 1.0.0
- [ x] I've checked [Known issues](https://github.com/hovancik/stretchly#known-issues)
- [ x] I've checked [Open issues](https://github.com/hovancik/stretchly/issues)

### Description
First of all thanks for great app:)

I've FR: to add Daily limit clock/counter - similar to microbreak and break - could we have a daily limit in hours? 
e.g. after 8hours it will display popup with message it's time to leave work. Might be very useful now especaily when many people have to work from home because of the covid-19.

Workrave does have it: https://mintguide.org/wp-content/uploads/posts/2014-03/1395392261_preferences.gif
Can Stretchly also have it?

Operating system: all

### Steps to Reproduce
N/a
**Expected behavior:**
After time's up (e.g. 8h) should display message it's time to stop working today.

**Actual behavior:**
no such feature

hovancik:
Hi @mwoz123 and thanks for the FR :) 

I guess we'll see in the future, but right now I have other things I would like to work on.

mwoz123:
Idea tech details ("moved" from #680) :

> Daily limit should have same/similar rule set as micro- and break.
> 
> Popup should appear after (default) e.g. 6 or 8h, and can be postpone or skipped.
> Daily limit should have no "break duration" time.
> 
> Once breaks are paused, or rested - daily limit counter should also be paused/rested.
> (I havne't check how "natural break" works but it would be good to delay daily limit after certain period of inactivity or even reset after long enough inactivity - but I think this should be next PR)
> 
> I created "skip to daily limit" - it doesnt' have much logic sense - it's temp function to simplify development, so don't be confused by that;)
> Should be removed once everything is working fine.

@hovancik answer:

> To be honest, I am not sure about this. I don't like the idea of adding features that other apps have, Workrave in this case. I would prefer Stretchly to be different.
> 
> Stretchly is break time reminder app. This feature looks like a timer to finish ones shift. There are hundreds of apps for counting time or timers. I would prefer to keep Stretchly simple, as every extra features make it harder to maintain and to add core features. This feature seems to be as it should be separate application.

my response:

> > I would prefer to keep Stretchly simple, as every extra features make it harder to maintain and to add core features.
> 
> (From developer point of view) I totally agree.
[...]

PR #680, opened 2020-09-09
w.i.p. daily limit counter

mwoz123:
Issue: #667 

### Requirements

### Description of the Change

Help adapting lines:
https://github.com/hovancik/stretchly/pull/680/files#diff-0a3ec7d7c0484270905e25a072b4f0c8L74
apprecaited. I don't quite follow the logic here (L74-L102) :/

Daily limit counter - informs when it's time to stop working with computer

Zrzut ekranu z 2020-09-09 21-38-39

Zrzut ekranu z 2020-09-09 21-38-07


hovancik:
Hey @mwoz123 , we should first talk about feature before writing any code. 

How do you imagine this should work?

mwoz123:
Daily limit should have same/similar rule set as micro- and break. 

Popup should appear after (default) e.g. 6 or 8h, and can be postpone or skipped.
Daily limit should have no "break duration" time.

Once breaks are paused, or rested - daily limit counter should also be paused/rested.
(I havne't check how "natural break" works but it would be good to delay daily limit after certain period of inactivity or even reset after long enough inactivity - but I think this should be next PR)

I created "skip to daily limit" - it doesnt' have much logic sense - it's temp function to simplify development, so don't be confused by that;)
Should be removed once everything is working fine.

hovancik:
To be honest, I am not sure about this. I don't like the idea of adding features that other apps have, Workrave in this case. I would prefer Stretchly to be different. 

Stretchly is **break time reminder app**. This feature looks like a **timer to finish ones shift**. There are hundreds of apps for counting time or timers. I would prefer to keep Stretchly simple, as every extra features make it harder to maintain and to add core features. This feature seems to be as it should be separate application.

mwoz123:
> I would prefer to keep Stretchly simple, as every extra features make it harder to maintain and to add core features.

(From developer point of view) I totally agree.
Especially when there probably would be a need to change [breakPlanner.js lines 74-102](https://github.com/hovancik/stretchly/blob/7702f9c1c98f632ad7f0def4c2afdf18558c824d/app/breaksPlanner.js#L87) which already has complicated logic.

>There are hundreds of apps for counting time or timers

Here I disagree. 
Why someone should have 2 counting down time apps and displaying notification once time is up running in same time on same system?

How about creating array that will contain number of break timers (and their configuraiton) and if someone will want to have 10 time's up reminders (instead 2 - but I belive most users will want at most 3) can be easily set in configuration. If some wants 1 will change the configuration to 1, and so on.. 
How about that?

>Stretchly is break time reminder app. 

Btw I was wondering why it's not marked with "RSI" or "prevent ["Repetitive strain injury" (RSI)](https://www.nhs.uk/conditions/repetitive-strain-injury-rsi/) in tags or description. 
It might be easier to find this app;) 
and wanted to propose it to you;)

Anyway is there a chance such feature will be accepted? I thought so from your anser https://github.com/hovancik/stretchly/issues/667#issuecomment-683282721

hovancik:
I will need to think more. Because much more people request that I will add "custom working times", which goes in similar direction but it would be then close to impossible to have both. Maybe a combination where people can define working hours and be notified at the end?

mwoz123:
"moving" conversation to #667

mwoz123:
@hovancik  I replied in issue :
- https://github.com/hovancik/stretchly/issues/667#issuecomment-748940100
- https://github.com/hovancik/stretchly/issues/667#issuecomment-748925544

What you think?

PR #700, opened 2020-09-21
Control Stretchly from commandline

### LINKED ISSUE #458: Allow stretchly operations to be issued via commandline
### Prerequisites

- [x] I'm using latest version: v0.20.1

### Description

There is currently no way to "Reset Breaks" (for example) via command line. This would be helpful in multi-platform or multi-system environments (I use Linux and macOS side-by-side via Synergy), to enable automatic syncing of multiple stretchly instances.

### Steps to Reproduce

n/a

### Additional Information

n/a

papey:
Issue: closes #458 

### Requirements

### Description of the Change

Contribute and finish the work started on PR #567.

### Verification Process

I've checked that commands goes from commandline call to main Stretchly instance

### Other information

none

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,13 +84,19 @@ Clicking the *Stretchly* icon in your tray area will display the current status 
 
 <img src="tray.png" height="140">
 
-*Stretchly* is monitoring your idle time, so when you are idle for 5 minutes, breaks will be paused until you return.  
+*Stretchly* is monitoring your idle time, so when you are idle for 5 minutes, breaks will be paused until you return.
 
 *Stretchly* is also monitoring Do Not Disturb mode, so breaks are paused when DnD mode is On.
 
 *Stretchly* follows the theme of your system and is also available in dark mode.
 
 <img src="dark.png" height="340">
+
+### Interact with stretchly from command line
+
+When a Stretchly instance is running, the `stretchly` command can be use to interact with it from the command line.
+
+Type `stretchly help` to get a list of all commands and options available as well as some examples.
 
 ## Preferences
 
@@ -112,7 +118,7 @@ To make sure that all works as expected, it's always good idea to check that for
 
 Some of the extra settings are available in Contributor Preferences for [Contributors](#contributor-preferences). Those are marked by [![Contributor Preferences](https://img.shields.io/badge/Contributor_Preferences-✔-success)](#contributor-preferences) badge.
 
-**Note:** Before 1.0, Mini Breaks and Long Breaks were called Microbreaks and Breaks, respectively. To keep the upgrade smooth they still use that name in preferences file and in code.  
+**Note:** Before 1.0, Mini Breaks and Long Breaks were called Microbreaks and Breaks, respectively. To keep the upgrade smooth they still use that name in preferences file and in code.
 
 #### Editing Break ideas
 In the preferences file, change `useIdeasFromSettings: false,` to `useIdeasFromSettings: true,` and edit `breakIdeas` and `microbreakIdeas`.
@@ -151,7 +157,7 @@ In the preferences file, set `volume` to your preferred value. Default value is 
 In the preferences file, set `endBreakShortcut` to your preferred value. We do not validate this input, so please check [Electron's documentation](https://www.electronjs.org/docs/api/accelerator) for available values for key and modifier.
 
 #### Appearance [![Contributor Preferences](https://img.shields.io/badge/Contributor_Preferences-✔-success)](#contributor-preferences)
-In the preferences file, change `themeSource: 'system'` to either `'light'` or `'dark'` to always use the specified theme.  
+In the preferences file, change `themeSource: 'system'` to either `'light'` or `'dark'` to always use the specified theme.
 
 #### Welcome window [![Contributor Preferences](https://img.shields.io/badge/Contributor_Preferences-✔-success)](#contributor-preferences)
 To show the Welcome window again on the next start, change `"isFirstRun"` to `true`.
@@ -317,7 +323,7 @@ Sounds used in this application are listed [here](http://freesound.org/people/ho
 - `silence` by [parcodeisuoni](http://freesound.org/people/parcodeisuoni/), available under the [Attribution License](http://creativecommons.org/licenses/by/3.0/).
 
 #### Fonts credits
-This app uses [Open Sans](https://fonts.google.com/specimen/Open+Sans) fonts licensed under the [Apache License, Version 2.0](http://www.apache.org/licenses/LICENSE-2.0).  
+This app uses [Open Sans](https://fonts.google.com/specimen/Open+Sans) fonts licensed under the [Apache License, Version 2.0](http://www.apache.org/licenses/LICENSE-2.0).
 
 ## License
 See [LICENSE](https://github.com/hovancik/stretchly/blob/master/LICENSE) file.
--- a/app/main.js
+++ b/app/main.js
@@ -17,6 +17,7 @@ const IdeasLoader = require('./utils/ideasLoader')
 const BreaksPlanner = require('./breaksPlanner')
 const AppIcon = require('./utils/appIcon')
 const { UntilMorning } = require('./utils/untilMorning')
+const Command = require('./utils/commands')
 
 let microbreakIdeas
 let breakIdeas
@@ -32,6 +33,7 @@ let syncPreferencesWindow = null
 let myStretchlyWindow = null
 let settings
 let pausedForSuspendOrLock = false
+let nextIdea = null
 
 app.setAppUserModelId('net.hovancik.stretchly')
 
@@ -43,7 +45,10 @@ global.shared = {
 const gotTheLock = app.requestSingleInstanceLock()
 
 if (!gotTheLock) {
-  console.log('Stretchly is already running.')
+  console.log('Stretchly command instance: started\n')
+  const args = process.argv.slice(app.isPackaged ? 1 : 2)
+  const cmd = new Command(args, app.getVersion())
+  cmd.runOrForward()
   app.quit()
   return
 }
@@ -52,6 +57,7 @@ app.on('ready', startProcessWin)
 app.on('ready', loadSettings)
 app.on('ready', createTrayIcon)
 app.on('ready', startPowerMonitoring)
+app.on('second-instance', runCommand)
 app.on('window-all-closed', () => {
   // do nothing, so app wont get closed
 })
@@ -384,7 +390,8 @@ function startMicrobreak () {
   const modalPath = `file://${__dirname}/microbreak.html`
   microbreakWins = []
 
-  const idea = settings.get('ideas') ? microbreakIdeas.randomElement : ['']
+  const idea = nextIdea || (settings.get('ideas') ? microbreakIdeas.randomElement : [''])
+  nextIdea = null
 
   if (settings.get('microbreakStartSoundPlaying') && !settings.get('silentNotifications')) {
     processWin.webContents.send('playSound', settings.get('audio'), settings.get('volume'))
@@ -492,7 +499,9 @@ function startBreak () {
   const modalPath = `file://${__dirname}/break.html`
   breakWins = []
 
-  const idea = settings.get('ideas') ? breakIdeas.randomElement : ['', '']
+  const defaultNextIdea = settings.get('ideas') ? breakIdeas.randomElement : ['', '']
+  const idea = nextIdea.map((val, index) => val || defaultNextIdea[index])
+  nextIdea = null
 
   if (settings.get('breakStartSoundPlaying') && !settings.get('silentNotifications')) {
     processWin.webContents.send('playSound', settings.get('audio'), settings.get('volume'))
@@ -911,6 +920,61 @@ function showNotification (text) {
     text: text,
     silent: settings.get('silentNotifications')
   })
+}
+
+function runCommand (event, argv, workingDirectory) {
+  const args = argv.slice(app.isPackaged ? 1 : 2)
+  const cmd = new Command(args, app.getVersion())
+
+  // if this command is already executed by the second-instance, return early
+  if (!cmd.checkInMain()) {
+    log.info('Stretchly: command executed in second-instance, dropped in main instance')
+    return
+  }
+
+  switch (cmd.command) {
+    case 'reset':
+      log.info('Stretchly: reseting breaks (requested by second instance)')
+      resetBreaks()
+      break
+
+    case 'mini':
+      log.info('Stretchly: skip to Mini Break (requested by second instance)')
+      if (cmd.options.title) nextIdea = [cmd.options.title]
+      if (!cmd.options.noskip) skipToMicrobreak()
+      break
+
+    case 'long':
+      log.info('Stretchly: skip to Long Break (requested by second instance)')
+      nextIdea = [cmd.options.title ? cmd.options.title : null, cmd.options.text ? cmd.options.text : null]
+      if (!cmd.options.noskip) skipToBreak()
+      break
+
+    case 'resume':
+      log.info('Stretchly: resume Breaks (requested by second instance)')
+      if (breakPlanner.isPaused) resumeBreaks(false)
+      break
+
+    case 'toggle':
+      log.info('Stretchly: toggle Breaks (requested by second instance)')
+      if (breakPlanner.isPaused) resumeBreaks(false)
+      else pauseBreaks(1)
+      break
+
+    case 'pause':
+      log.info('Stretchly: pause Breaks (requested by second instance)')
+      var ms = cmd.durationToMs(settings)
+      // -1 indicates an invalid value
+      if (ms === -1) {
+        log.error('Stretchly: error when parsing duration to ms because of unvalid value')
+        return
+      }
+      pauseBreaks(ms)
+      break
+
+    default:
+      log.error(`Stretchly: Command ${cmd.command} is not supported`)
+  }
 }
 
 ipcMain.on('postpone-microbreak', function (event, shouldPlaySound) {
--- a/app/utils/commands.js
+++ b/app/utils/commands.js
@@ -1,0 +1,249 @@
+const { UntilMorning } = require('./untilMorning')
+const log = require('electron-log')
+
+const allOptions = {
+  title: {
+    long: '--title',
+    short: '-T',
+    description: 'Specify title for next break (Long or Mini)',
+    withValue: true
+  },
+  text: {
+    long: '--text',
+    short: '-t',
+    description: 'Specify text for next break (Long Break only)',
+    withValue: true
+  },
+  noskip: {
+    long: '--noskip',
+    short: '-n',
+    description: 'Do not skip directly to this break (Long or Mini)',
+    withValue: false
+  },
+  duration: {
+    long: '--duration',
+    short: '-d',
+    description: 'Specify duration for pausing breaks (Pause only) [infinitely|until-morning|seconds (as number)]',
+    withValue: true
+  }
+}
+
+const allCommands = {
+  help: {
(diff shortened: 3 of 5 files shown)


papey:
Looking at the code, I see mainly two issues we need to discuss in order to get things done.

First things first, for the implementation part, after playing with multiple kind of inputs, I think this approach : https://github.com/hovancik/stretchly/pull/567#issuecomment-634654738 is the best one. It's simple and efficient, without the need to type too much extra chars.

The other one is about how commands should be handled.

For now, we are only passing args to the main instance using : 

```app.on('second-instance', runCommand)``` 

Using the `second-instance` event, `stdout` is bind to the main instance so you can't handle commands that needs to return outputs to the shell launching the command. So what about the `help` and version `command` ?

One simple idea is to break command handling into two parts, the ones that needs to be forwarded to the main instance and the one that only outputs text to the user. With this solution, the ugly thing will be : 

* the need to parse args twice, first time in the `comandline instance` and another time in the `main instance`, in the callback
* some commands will go into the `second-instance` callback but will do nothing (`help` and `version` at least)

Thank you in advance for your highlights and guidance on this subjects.

**Comment by hovancik** [comment]
:+1: for the interface you chose. I think when testing, you'll see if it feels natural or not. 

Regarding help and version: when I was playing with code first time around, I came to the same conclusion as you, that there should have to be 2 parsers. I haven't found other way that time.

**Comment by papey** [comment]
Consider this as WIP, i'm trying to find the best way to handle this but a lib might be a good option.

**Comment by papey** [comment]
Here is a the help message for this first draft

```
Stretchly is already running, sending command

Usage: stretchly <command> [options]

Commmands:
        stretchly help            Show this help message
        stretchly version         Show current stretchly version
        stretchly reset           Reset breaks
        stretchly pause [options] Pause breaks
        stretchly resume          Resume from a pause
        stretchly toggle          Toggle breaks between resume/paused
        stretchly mini [options]  Skips to next Mini Break
        stretchly long [options]  Skips to next Long Break

Options:
        -T, --title    Specify title for next break (Long or Mini)
        -t, --text     Specify text for next break (Long Break only)
        -s, --skip     Skip directly to this break (Long or Mini)
        -d, --duration Specify duration for pausing breaks (Pause only)
```

For now, everything is done without any parsing lib. I've made some assumptions in order to keep things simple in the command handling code since we need to handle commands twice (one in the `second instance` one in the `main instance`).

All the important stuff is in a dedicated module `commands.js`. Even if it's quite simple, I've tried to organize all the things logically.

The most complex code is in the help message. This is a generated message from all the commands and associated options defined in the module. Handling the padding to get a nice output is what makes things a bit ugly in this code. It needs a bit of rework but it does the job for now.

I also add a `toggle` command since this is the one I need as keyboard shortcut.

File: app/main.js [hunk 1]
```diff
@@ -41,7 +43,10 @@ global.shared = {
 const gotTheLock = app.requestSingleInstanceLock()
 
 if (!gotTheLock) {
-  console.log('Stretchly is already running.')
+  console.log('Stretchly is already running, sending command\n')
```
hovancik, line 46: I've recently added `electron-log` package for logging. I think we could try using it for logging this as well. 

I've been using this way of logging: https://github.com/hovancik/stretchly#logging Maybe use something like `Stretchly 2nd instance: message` or sth like that. I guess it should go to same file.

app/utils/commands.js
```diff
@@ -0,0 +1,241 @@
...
+    this.supported = allCommands
+
+    this.parse(input)
+  }
+
+  parse (input) {
+    this.command = input[0]
+
+    if (!this.supported[this.command]) {
+      throw new Error(`Error: command ${this.command} is not supported`)
```
hovancik, line 90: I usually run with `npm run dev`. Could we make it run fine with this? 

```
"dev": "cross-env NODE_ENV=development electron ./app --enable-logging --remote-debugging-port=9222",
```

It's throwing `Error: Error: command --enable-logging is not supported` now.

app/utils/commands.js
```diff
@@ -0,0 +1,241 @@
...
+            options[opt.long.slice(2)] = opts[i + 1]
+            i++
+          } else {
+            options[opt.long.slice(2)] = true
+          }
+        }
+      })
+
+      if (!valid) {
+        throw new Error(`Error: options ${name} is not valid for command ${this.command}`)
```
hovancik, line 120: Do we need to throw here? I don't think we should get window from command line 

image

papey: It's something I wanted to discuss with you, I left this on purpose in order to get a clear error. Since it's in the second instance, there is no way to get clear feedback. I can log it to stdout in the main instance. Another solution would be a notification, maybe ?

app/utils/commands.js
papey: It's thrown if you have another primary stretchly instance running. The code pass in https://github.com/papey/stretchly/blob/features/cmdline-automation/app/main.js#L45 this path. If you need multiple stretchly instance, this `gotTheLock` might not be a good option.

app/main.js
papey: See https://github.com/papey/stretchly/commit/554a61293032a69859973f50b6e5c52be373b4b2

app/utils/commands.js
papey: I replaced it with a log, see https://github.com/papey/stretchly/commit/554a61293032a69859973f50b6e5c52be373b4b2

app/main.js
hovancik: Awesome :) If you see some place that you think could be good one for some debug message, feel free to add it :)

app/utils/commands.js
hovancik: What I mean is: 

This runs fine:`npm start -- pause`
This does not run fine `npm run dev -- pause`

I guess we need to add `--enable-logging --remote-debugging-port=9222` to valid comands, but ignore them.

app/utils/commands.js
hovancik: I guess we should try to do as much of validation as possible in command instance, with errors to output. So that primary instance doesn't need to care. 

I am not sure for example what `checkInForward` is for in main.

papey: I renamed it and add some comments, it's the function run in the main instance to return early if the command is a `second-instance` kind (`version` and `help` atm).

papey: I fixed it naively here : https://github.com/hovancik/stretchly/pull/700/commits/42c3b2e440645432d68ea710d5d07abf38dadb4e

app/utils/commands.js
```diff
@@ -0,0 +1,253 @@
...
+    this.version = version
+    this.supported = allCommands
+
+    this.parse(input)
+  }
+
+  parse (input) {
+    // filter out electron flags first
+    let i = 0
+    while (input[i].startsWith('--')) {
```
hovancik, line 90: `npm run dev` gives error

```
TypeError: Cannot read property 'startsWith' of undefined
```

app/utils/commands.js
```diff
@@ -0,0 +1,253 @@
...
+      if (!valid) {
+        log.error(`Error: options ${name} is not valid for command ${this.command}`)
+      }
+    }
+
+    return options
+  }
+
+  runOrForward () {
+    if (!this.command) {
```
hovancik, line 140: I think we don't need this as we are making `help` the default command?

app/utils/commands.js
```diff
@@ -0,0 +1,253 @@
...
+      console.error('Error: command is undefined')
+      return
+    }
+
+    if (!this.supported[this.command]) {
```
hovancik, line 145: I think this will never happen as we catch this in `parse`. Let's use error there instead of throwing?

app/utils/commands.js
papey: Ok, it triggers when you have an already running instance, fixing it

app/utils/commands.js
papey: Fixed by https://github.com/hovancik/stretchly/pull/700/commits/c82e0d097aa1ab4d4a8c7ba6d60c2d07d7d7c223

app/utils/commands.js
papey: Fixed by https://github.com/hovancik/stretchly/pull/700/commits/c82e0d097aa1ab4d4a8c7ba6d60c2d07d7d7c223

app/utils/commands.js
papey: Fixed by https://github.com/hovancik/stretchly/pull/700/commits/1424058ee0283fa2dbab94784456718c93f8908a

app/main.js
```diff
@@ -915,6 +923,47 @@ function showNotification (text) {
...
+  if (!cmd.checkInMain()) return
+
+  switch (cmd.command) {
+    case 'reset':
+      resetBreaks()
+      break
+
+    case 'mini':
+      if (cmd.options.title) nextIdea = [cmd.options.title, '']
+      if (cmd.options.skip) skipToMicrobreak(false)
```
hovancik, line 940: `skipToBreak` and `skipToMicrobreak` don't have any params

app/main.js
```diff
@@ -915,6 +923,47 @@ function showNotification (text) {
...
 
+function runCommand (event, argv, workingDirectory) {
+  const args = argv.slice(app.isPackaged ? 1 : 2)
+  const cmd = new Command(args, app.getVersion())
+
+  // if this command is already executed by the second-instance, return early
+  if (!cmd.checkInMain()) return
+
+  switch (cmd.command) {
+    case 'reset':
```
hovancik, line 934: Can we `log.info` for each option? So in case I am debugging something I know  where events come from

app/main.js
```diff
@@ -915,6 +923,47 @@ function showNotification (text) {
...
+      else pauseBreaks(1)
+      break
+
+    case 'pause':
+      var ms = cmd.durationToMs(settings)
+      pauseBreaks(ms)
+      break
+
+    default:
+      log.Error(`Unsupported command ${cmd.command}`)
```
hovancik, line 963: `Error` should be lower-case, I guess.

app/main.js
papey: https://github.com/papey/stretchly/commit/f7df5cb82d9df19aed4e12d846aa48d2fae0285c

app/main.js
papey: https://github.com/papey/stretchly/commit/470ba06faafad669674b590b48f08e5ae0e504d8

app/main.js
papey: https://github.com/papey/stretchly/commit/1e7b00595434ce69819118b9063385bdd1bea32a

hovancik:
So I believe I am pretty happy with the code and I understand what it does :) But as always with me, some things: 

1. Do you feel like writing tests? I imagine `Command` could be tested :) 
2. Have you tried with packaged version of app? I wonder whether `electron-log` outputs are visible, then. Asking because I wonder whether we should use `electron-log` or node's `log` in Command. On one hand, we would have it in logs, on the other, it's a bit uglier output (maybe that's controllable?). Right now we have a combination + Throw. 
3. Could we get rid of throw? One usage in Parse is not an issue, I think, as we are parsing from second instance, so if we would log, user will see an error in output. The other usage in `durationToMs`is harder, as it seems to be called in main instance. Could we check the value maybe when parsing? What do you think? 
4. The last think is readme, I guess only mentioning one example + saying no command is `help` would be enough? 

Let me know what you think.

papey:
1. Now that we have a somewhat solid base that we will not trash away, for sure I will add test to it !

2. I will test that ASAP, for me, we have to `electron-log` in the "electron app" ie, the main instance and use node's console log in the second instance.

3. In both case, the command is parse in both the main instance and the second instance, in both cases a logging line could be enough (?), I think.

4. For the README. I think a simple mention saying that you can interact with an already started and use `stretchly help` to get a list of commands should be enough.

hovancik:
I agree on all points :) :+1:

papey:
I've updated my comment to point to changes we discuss. The only thing left is testing in the packaged app. Will do it ASAP.

app/main.js
```diff
@@ -43,7 +45,10 @@ global.shared = {
 const gotTheLock = app.requestSingleInstanceLock()
 
 if (!gotTheLock) {
-  console.log('Stretchly is already running.')
+  console.log('Stretchly command instance: started\n')
```
hovancik, line 48: Only wondering: why do you use `\n`? :)

papey: Because I find not quite redeable when help is printed : 

```
Stretchly command instance: started
Usage: stretchly <command> [options]

Commands:
	stretchly help            Show this help message
	stretchly version         Show current stretchly version
	stretchly reset           Reset breaks
	stretchly pause [options] Pause breaks
	stretchly resume          Resume from a pause
	stretchly toggle          Toggle breaks between resume/paused
	stretchly mini [options]  Skips to next Mini Break
	stretchly long [options]  Skips to next Long Break

Options:
	-T, --title    Specify title for next break (Long or Mini)
	-t, --text     Specify text for next break (Long Break only)
	-s, --skip     Skip directly to this break (Long or Mini)
	-d, --duration Specify duration for pausing breaks (Pause only) [infinitely|until-morning|seconds (as number)]

Examples:
	stretchly pause                                    Pause breaks indefinitly
	stretchly pause -d 3600                            Pause breaks for one hour
	stretchly mini -T "Stretch up !" --skip            Skips to next Mini Break with "Stretch up!" title
	stretchly long -T "Stretch up !" -t "Go stretch !" Set next long break title to "Stretch up !" and text to "Go stretch !"
```

vs

```
Stretchly command instance: started
Usage: stretchly <command> [options]

Commands:
	stretchly help            Show this help message
	stretchly version         Show current stretchly version
	stretchly reset           Reset breaks
	stretchly pause [options] Pause breaks
	stretchly resume          Resume from a pause
	stretchly toggle          Toggle breaks between resume/paused
	stretchly mini [options]  Skips to next Mini Break
	stretchly long [options]  Skips to next Long Break

Options:
	-T, --title    Specify title for next break (Long or Mini)
	-t, --text     Specify text for next break (Long Break only)
	-s, --skip     Skip directly to this break (Long or Mini)
	-d, --duration Specify duration for pausing breaks (Pause only) [infinitely|until-morning|seconds (as number)]

Examples:
	stretchly pause                                    Pause breaks indefinitly
	stretchly pause -d 3600                            Pause breaks for one hour
	stretchly mini -T "Stretch up !" --skip            Skips to next Mini Break with "Stretch up!" title
	stretchly long -T "Stretch up !" -t "Go stretch !" Set next long break title to "Stretch up !" and text to "Go stretch !"
```

But it's a matter of taste

hovancik: Ah, ok, you're right :)

papey:
It everything looks good to you I will finish the last detail, re-test all the stuff for a day or two, update CHANGELOG and squash everything !

README.md
```diff
@@ -92,6 +92,12 @@ Clicking the *Stretchly* icon in your tray area will display the current status
 
 <img src="dark.png" height="340">
 
+## Interact with stretchly from command line
```
hovancik, line 95: can we make this h3? `###`

hovancik:
Looks good to me! 

I'll try dev version as well, but I trust that if you use it for your purposes it will be well tested. Let me know if packaged version is OK.

papey:
@hovancik just to be shure, is `npm run pack` and then running stretchly from `./dist/linux-unpacked/stretchly` enougth for testing the packaged version ? Sorry if there is some documentation about it, I did not find it.

hovancik:
Good question, not sure. What Linux do you run? I would edit `package.json` to only build your distribution, then run `npm run dist` and install as you would with normal installer.

papey:
Hey @hovancik, after a week of usage, I've rebase the branch against master, fixed a bug / weird behavior with the `long` command (see https://github.com/hovancik/stretchly/pull/700/commits/32aabbc8e769e161fedccd8bde283b24c89bbe5f). I also changed the default behavior of `long` and `mini` command. I prefer them to skip by default so `skip` option becomes `noskip`. Packaged version works for me. We should be close to squash and merge.

hovancik:
@papey looks good to me :) only think left: 
```
-s, --noskip   Do not skip directly to this break (Long or Mini)

```
should the short one be `-n`, or something else?

papey:
Was note sure about this so I left this but yeah I think you're right

hovancik (approved):

papey:
All squashed ! :+1:

hovancik:
Thanks! A lot of work went into this :)

papey:
Thanks a lot for the feedback, was a pleasure :+1:

decadent:
Sorry for barging in belatedly, but it just popped up in my mind that I should take a look at the pausing command (my main interest here), and: perhaps the duration should specify minutes if it's a number? I'm very much sure that people don't usually think of work/leisure intervals in seconds. I actually have a timer app with a CLI interface (for Alfred), and I call it with the argument in minutes, which works splendidly. The list of my pause intervals in Stretchly even has a ‘30 minutes’ option, and that sounds way more natural than ‘1800 seconds’.

Personally I would also parse values in the formats ‘2h’ and ‘2h30m’, for they are also very convenient to specify on the command line. Parsing them is straightforward with regular expressions, it's basically two lines of code—I can write it if you wish. Since the commands themselves are in English, I guess the format also doesn't need internationalization.

app/utils/commands.js
```diff
@@ -0,0 +1,249 @@
...
+  noskip: {
+    long: '--noskip',
+    short: '-n',
+    description: 'Do not skip directly to this break (Long or Mini)',
+    withValue: false
+  },
+  duration: {
+    long: '--duration',
+    short: '-d',
+    description: 'Specify duration for pausing breaks (Pause only) [infinitely|until-morning|seconds (as number)]',
```
decadent, line 26: ‘infinitely’ here should probably be ’indefinitely’, since that's what used later in the code and in the app's menu.

app/utils/commands.js
```diff
@@ -0,0 +1,249 @@
...
+const allExamples = [{
+  cmd: 'stretchly pause',
+  description: 'Pause breaks indefinitly'
+},
+{
+  cmd: 'stretchly pause -d 3600',
+  description: 'Pause breaks for one hour'
+},
+{
+  cmd: 'stretchly mini -T "Stretch up !"',
```
decadent, line 70: A little nitpick: perhaps the command examples shouldn't have spaces before exclamation marks—like there's no spaces in the descriptions.

app/utils/commands.js
```diff
@@ -0,0 +1,249 @@
...
+  },
+  long: {
+    description: 'Skips to and customize next Long Break',
+    options: [allOptions.text, allOptions.title, allOptions.noskip]
+  }
+}
+
+const allExamples = [{
+  cmd: 'stretchly pause',
+  description: 'Pause breaks indefinitly'
```
decadent, line 63: A typo: it should be ‘indefinitely’.

papey:
Sure it's doable. @hovancik if you want I can make a quick PR to integrate features requested by @decadent before next release

app/utils/commands.js
papey: Sorry, french habits.

app/utils/commands.js
papey: Yeah sure, just a typo.

app/utils/commands.js
papey: Nice catch

papey:
A starting block is available here https://github.com/papey/stretchly/tree/features/improve-pause-cmd.

hovancik:
Sounds good :)

hovancik:
@papey I'm also getting this, when Long Break is about to start: 

image

papey:
Thanks for feedback will add a quick fix

papey:
Since this is closed, discussion is moved to https://github.com/hovancik/stretchly/pull/748

Issue #704, opened 2020-09-25
Korean Translation Has a Few Kinks


hcsong213:
### Prerequisites

- [x] I'm using latest version: v 1.1.0 
- [x] I've checked [Known issues](https://github.com/hovancik/stretchly#known-issues)
- [x] I've checked [Open issues](https://github.com/hovancik/stretchly/issues)

### Description

Korean translation for "skip to next break" is awkward. Current phrasing is more similar to "get/receive next break".

Beside this, this Issue also serves to express my general interest in refining and expanding the Korean translation.

Operating system:  Windows 10

### Steps to Reproduce

1. Click on stretchly icon in taskbar
2. See the button for "skip to next break" translated as "쉬는 시간 갖기" (갖기 is not the best phrasing here)

**Expected behavior:** Said button translated as "다음 쉬는시간 으로" (translates roughly as "to next break"; more natural in my and my parents' opinions) 

**Actual behavior:** Said button translated as "쉬는 시간 갖기" 

**Reproduces how often:** 100% 

### Additional Information

To briefly state my relationship with Korean, I immigrated from South Korea and use Korean in my house daily. I frequently use Korean websites which include UI elements similar to the one discussed in this Issue.

There are other elements of the translation that is culturally off-putting.

By no means am I trying to sound derogatory. I appreciate the devs including so many languages so that more people could use this app. ❤️

hovancik:
Hi, feel free to create a PR with fix :)

hcsong213:
Can I also translate other components that haven't been translated yet?

hovancik:
Sure :) There are always new features so translations go out of date easy

PR #707, opened 2020-09-26
Updates to Korean Translation

hcsong213:
Issue: #704 

### Requirements

*I used node v.12.18.1 (json file asks for 12.14.1)
I'm on a windows machine, and nvm analogs to Windows are much less elegant and more error prone. Because I'm working on a project in my startup which uses node, I didn't want to risk any complications.
Since I only edited the ko.json file , it should not be a problem.*
The only new code is changes in .json file

### Description of the Change

Updated Korean translations (in ko.json) for more translated elements, better spacing, and better cultural fit.

### Verification Process

- How did you verify that all new functionality works as expected?
By running the app with the new source code using `npm start`
- How did you verify that all changed functionality works as expected?
Same as above.
- How did you verify that the change has not introduced any regressions?
By using the app and perceiving that no undesired side effects arose.

### Other information
When opening the markdown files on vscode, my prettier formatter added a bunch of random spaces. While I don't think that this will break the file, I'll be more than willing to rebase and revert the files to their original state.


hovancik (approved):
:kr: thanks a lot!

PR #746, opened 2020-10-24
Fixed eye distance in microbreak

AntonyGarand:
Issue: NA

### Requirements

- [ ]  all new code has tests to ensure against regressions. [ Irrelevant ]
- [X]  after PR is approved, all commits in it are [squashed](https://gitbetter.substack.com/p/how-to-squash-git-commits)

### Description of the Change

Replaced the `20 feet` with `20 meters` to be consistent with the other references to the eye focus distance:   
https://github.com/hovancik/stretchly/blob/master/app/utils/defaultSettings.js#L226 and  
https://github.com/hovancik/stretchly/blob/master/app/utils/defaultMicrobreakIdeas.js#L47

### Verification Process

Text change, didn't verify anything

### Other information

I'd appreciate if you could add the `hacktoberfest-accepted` label so it counts towards my valid PR's for the event 😄

### DIFF
--- a/app/utils/defaultMicrobreakIdeas.js
+++ b/app/utils/defaultMicrobreakIdeas.js
@@ -88,7 +88,7 @@ module.exports = [
     enabled: true
   },
   {
-    data: 'One should focus every 20 minutes for 20 seconds on an object that is 20 feet distance.',
+    data: 'One should focus every 20 minutes for 20 seconds on an object that is at least 20 meters away.',
     enabled: true
   },
   {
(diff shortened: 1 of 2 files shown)


hovancik:
Hi @AntonyGarand , I don't see any reason why this should be changed. I think 20 feet are perfectly fine.

AntonyGarand:
I don't mind if feets or meters are kept, but I think both of the messages
should be consistent and use the same distance.
With one being "20 feets" and one being "20 meters", it's confusing which
one's the right way.

PR #768, opened 2020-11-13
Duplicated code

hovancik:
Issue: #

### Requirements

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -342,6 +342,51 @@ function startBreakNotification () {
   updateTray()
 }
 
+function generateWindowsOptions(localDisplayId) {
+    const windowOptions = {
+        width: Number.parseInt(displaysWidth(localDisplayId) * settings.get('breakWindowWidth')),
+        height: Number.parseInt(displaysHeight(localDisplayId) * settings.get('breakWindowHeight')),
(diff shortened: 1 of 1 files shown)


app/main.js
```diff
@@ -301,13 +281,7 @@ function createContributorSettingsWindow () {
...
 }
 
-function createSyncPreferencesWindow () {
-  if (syncPreferencesWindow) {
-    syncPreferencesWindow.show()
-    return
-  }
-
-  const syncPreferencesUrl = 'https://my.stretchly.net/app/v1/sync'
+function newStretchlyWindow (syncPreferencesUrl) {
```
hovancik, line 284: I don't think we should do this. `syncPreferencesWindow` and `myStretchlyWindow` might look similar (in behavior and params) now but it needn't need to be in future. 

You are also mixing them together: in `newStretchlyWindow` you use `syncPreferencesWindow` and then do `myStretchlyWindow = newStretchlyWindow(myStretchlyUrl)`. We could make that a param but I don't think it's safe to assume they will stay the same. 
For example 
```
    width: 1000,
    height: 700,
```
will be changing based on how the website looks. We could also parametrize that but then, we end up the same as `BrowserWindow` itself.

Issue #769, opened 2020-11-15
Add contrast color theme - black text on white background


mwoz123:
### Prerequisites

- [X] I'm using latest version: git latest
- [X] I've checked [Known issues](https://github.com/hovancik/stretchly#known-issues)
- [X] I've checked [Open issues](https://github.com/hovancik/stretchly/issues)

### Description

Add contrast color theme - black text on white background - white background would be clearly visible also for E-ink monitors like Dasung.

Operating system: all

mwoz123:
could be simple as: 

> <input type="radio" value="#fffff" name="mainColor" **_data-font-color=_**"#000000" id="blackOnWhite">

hovancik:
Thanks :) I like that. I am superbusy with life right now but I think I would probably add new theme. WIll need some work on theme html and css

PR #776, opened 2020-11-24
black on white theme

mwoz123:
Screenshot from 2020-11-24 21-27-46

Screenshot from 2020-11-25 19-45-04

Issue: #
https://github.com/hovancik/stretchly/issues/769

### Requirements

- [x] `npm run lint` reports no offenses (on new code).

### Description of the Change

Add contrast color theme - black text on white background - white background would be clearly visible also for E-ink monitors like Dasung.

### Verification Process

Manul tests


mwoz123:
@hovancik - I did a change adding new theme.

I'd to do changes in places where code was already duplicated e.g. for `microbreak` and `break`.  
`codeclimate` marks that previously duplicated code as my change and blocks merging. 

@hovancik  - can you take a look at that? Please compare [my git changes/diff](https://github.com/hovancik/stretchly/pull/776/files) with codeclimate report

hovancik:
Hi @mwoz123 , don't care about codeclimate :) 

I am still busy with life but will try checking in coming days

Thanks for waiting!

hovancik:
Hi @mwoz123 , 

thanks for the PR :tada:  I'm trying it locally and have some feedback. In general, I would more logic to CSS (I use similar approch in few places, ie when making difference between Linux and macOS design):

- add class on top of html document (for example: https://github.com/hovancik/stretchly/blob/master/app/platform.js)
- then only change CSS where needed based on that (for example: https://github.com/hovancik/stretchly/blob/master/app/css/commons.css#L21)

You could for example add theme name as the class name and then we can update CSS based on that class. It would be simpler and easier to maintain. Plus the things I'll mention will be easier to fix :) 

Regarding the theme you added:
- icon is not visible here
image
- this need to be updated as well
image
- considering it's for E-ink, should we only use black and white colors there? 

That's for Break windows. 

For Preferences: 
- structure here needs to be updated (I am using grid to correctly position items https://github.com/hovancik/stretchly/blob/master/app/css/preferences.css#L123)
image

Let me know what you think, I'll add comments if I notice something else.

hovancik:
One more thing, here is how I managed to change colors of icons: https://hovancik.net/blog/2020/01/22/change-the-color-of-svg-image-via-css/

mwoz123:
@hovancik - thanks for review. But let me do a different proposal.

How about this :
https://www.geeksforgeeks.org/dynamic-styling-in-electronjs/
?

Solution: 
create theme folder, inside each theme will have separate css (not single css class per theme) where you can each element separately? 
Selected css will be dynamically loaded (as in the link above).

?

hovancik:
Hi and sorry again for late reply :) 

I am not sure if I see benefit of separating themes as they are themes: HTML is same, only colors are changing; so that's where I would handle that, in CSS. I don't plan to make themes so much different that it would make sense to do this as a optimization for future. 

I see you made some more commits, I'll try checking tomorrow and comment.

app/main.js
```diff
@@ -409,7 +409,7 @@ function startMicrobreak () {
       frame: false,
       show: false,
       transparent: settings.get('transparentMode'),
-      backgroundColor: calculateBackgroundColor(),
+      opacity: settings.get('transparentMode') ? settings.get('opacity') : 1,
```
hovancik, line 412: We can't do this as this is not multiplatform supported. That's why we used more complicated way with `calculateBackgroundColor()`

app/utils/defaultSettings.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
   microbreakPostponableDurationPercent: 30,
   breakPostponesLimit: 1,
   breakPostponableDurationPercent: 30,
-  mainColor: '#478484',
+  themeName: 'greenClouds',
```
hovancik, line 23: Please let's not change the names and values of settings. That's not something that I did even between version 0 and 1. I don't want users to loose settings... (And doing some more code to take care of that would be bad idea. )

app/main.js
mwoz123: ok

app/utils/defaultSettings.js
mwoz123: I don't get how you want to do it then:(

So/does? :
1. the mainColor var should stay and should still be used for setting background color?
2. new class should be at top of HTML to be able to theme it, but does it mean it should be theming everything except background color? 
3. if not using data-* attr on input for font color then I don't know how to do it

Could you clarify for me the solution you've on your mind? .

hovancik: Yes, we need `mainColor` as it also allows people so set whatever color they want for background. 

Currently, we have combination of colors in break windows that kind of works for every theme. (Not perfect, but it works.). With your new theme, we could keep most of the colors, but as I mentioned previously, some things were that good  on white (icon and yellow mouseover). So that's the minimum we need to update.  

Considering that you said you use this on e-ink, I thought we could make that new theme be only black and white colors. But we could do that later, or never :) I guess you can test and see what is enough. 

So I think easiest is to: 
- add class on top of break window (could be even the code of color, doesn't really matter)
- don't change anything for current themes
- when we have  your new theme, update CSS so that we have better colors for that tooltip and colors 

For example let's say your new theme is translated to new `123456` class on breaks:

```
.breaks > :nth-child(2) > a:hover {
  color: #FFFF00;
}

.123456 .breaks > :nth-child(2) > a:hover {
  color: #AAAA00;
}
```

This results in: 
- no change in how old themes works
- no change if users sets custom color for theme
- your new theme had different color on that link hover

So old CSS is not changed, you only kinda "duplicate" CSS that you want to "override" to make your theme better. It being font color, shadow, or anything that is set by CSS 

In future, we could then improve other themes in the same way and find better suiting colors for accessibility or so. 

You can move your "overrides" to specific file if you'd like. But I would not add it dynamically with JS, I would load it right away with the rest of CSS - it's not so huge file to worry about speed or so...

Does this makes sense?

mwoz123: 1. `when we have your new theme, update CSS so that we have better colors for that tooltip and colors`
do you mean "when we have your new theme **selected** (..)" ?

2. What you mean by `update CSS`? write js code -> if e-ink theme then break window set style ? add class?

3. `For example let's say your new theme is translated to new 123456 class on breaks:` how should I translate it? is it the solution from point 2?

mwoz123: 4. I'm wondering if you'd agree to reuse the current `mainColor` var + create new one e.g. fontColor (default val=white)  (selecting each theme will update those). 
We could reuse some coding done here and either:
- directly each time theme is selected in preferences change mainColor and fontColor variable,
- in UI create "custom theme" option at the bottom of all other themes and new inputs (in future to be replaced with color picker UI) for mainColor and font color (requires new var - customThemeEnabled(default=true, so prev users will keep their settings)) 
This way we should flexible create clean solution (without hacks in code that are often hard to maintain) and backwards compatible (as current configuration will work after those changes).

hovancik: OK, so let me try to explain differently what I mean. 

Some facts: 
- The new feature here is adding new Theme (Heaven White). 
- Theme means how Break Window is styled. 
- Old themes share same CSS (except background color). 
- CSS is across different files (we import `normalize.css`, `commons.css` and `break.css`)
- we choose theme to be used in Preferences window and save it as `#` value (this is what I meant by `selected`)

Adding new theme should be as easy as adding one more option in Preferences, but it is not in this case. 

Problem: Heaven White needs different colors for specific elements on Break Window. 

Solution: Leave current CSS as is, but add code for Heaven White  that will take the priority over current CSS in Break Window when Heaven White is current theme, (You can do that by using CSS class - (https://www.w3schools.com/css/css_specificity.asp) ). 

You are already adding new CSS class do achieve that here: 

```
document.body.classList.add(theme)
``` 
(We need to change it to use the old settings as I mentioned but that's detail: you can take `mainColor` from settings, remove `#` and set that as class to be added.)
 
With that change, we now have additional class on Break Window telling us what theme are we using. 

So for example font color is currently in `break.css`:

```
.breaks {
...
  color: #FFFFFF;
...
}
```
If we added `FFFFFF` class for Heaven White, in CSS you just do (or similar depending where you put the class): 

```
.FFFFFF .breaks {
...
  color: #000000;
...
}

```
This should override the `.breaks` selector. right? 

It's pretty much the same what you did in your new file `themes.css` but there is no need to specify all previous ones, you can just do override for new theme. As example: when user did not change theme in Preferences, `478484` class would be added to Break window. But that does not change the old behaviour, so  all good. 

Therefore there is no need to save font-color in preferences. The same way you can override any other theming changes needed. 

This also allows for later gradual changes for current themes if there's need to make them better.  

We could of course add `fontColor` value to settings to allow more configuration for users, but that's for future. I would like to keep PRs simple, so only one feature at time. Also, adding font color would not be enough, though, as we also have the need for mouse-over color etc.. I a not sure how theming will work in future...

In general, I would prefer to keep config simple, too much ogf a choice makes bad UX.

mwoz123:
Thanks for explanation. I thought you want to something more complex. 
Canceling this PR. Created new https://github.com/hovancik/stretchly/pull/808

hovancik:
Thanks! Will take a look latest over weekend

Issue #868, opened 2021-02-21
Add Latvian language


darksworm:
### Prerequisites

- [x] I'm using latest version: v1.4.0
- [x] I've checked [Known issues](https://github.com/hovancik/stretchly#known-issues)
- [x] I've checked [Open issues](https://github.com/hovancik/stretchly/issues)

### Description

Please add the latvian language to stretchly as it has been recently translated.

PR #869, opened 2021-02-21
add Latvian language

darksworm:
Issue: #868

### Requirements

- [ ] `npm run test` is error-free.  - **it actually fails but not due to these changes.**

### Description of the Change

Added latvian translations JSON file from weblate and added Latvian to the language selector in settings.

### Verification Process

After making the changes I walked through the application (yes, also the contributor settings) and re-read all the translations to make sure they are OK. 

Some are not perfect, due to how word ending formation differs in Latvian vs English (they are very context dependent in Latvian and differ a lot).

For example in english "for 10 minutes", "about 10 minutes" uses the same form of the word "minute". In Latvian, those become "uz 10 minūtēm", "aptuveni 10 minūtes" - having these forms correct in Stretchly is currently not possible.


hovancik:
Hi @darksworm , I've merged your changes from Weblate here: #867. They have integration, so no need to include JSON here. You should also add the language option to Welcome.html. I use flags from here: https://github.com/HatScripts/circle-flags

If you feel that some other form would be better in Latvian, feel free to change it as you see fit. I do the same when translate to Slovak.

darksworm:
> Hi @darksworm , I've merged your changes from Weblate here: #867. They have integration, so no need to include JSON here. You should also add the language option to Welcome.html. I use flags from here: https://github.com/HatScripts/circle-flags
> 
> If you feel that some other form would be better in Latvian, feel free to change it as you see fit. I do the same when translate to Slovak.

Really cool that weblate creates pull requests automatically! 

I've removed the JSON file and added the Latvian language option to `welcome.html`.
image

hovancik (approved):

Issue #437, opened 2019-07-15
stretchly should refocus to itself if someone escapes to different app with ctrl/cmd+tab or with another way


cattrace:
## enviroment
mac 2017 pro touchbar
stretchly 0.20.0  fullscreen mode

## description
![image](https://user-images.githubusercontent.com/4670343/61203786-fc485600-a71d-11e9-8316-ab6141512b93.png)

when you use Alfred with shotcut   , and then type another application(like chrome) , mac will switch that application (at another desktop) ,then you escaped.

hovancik:
Yeah ;/ 

Stretchly is a regular window, so there will be always a way to get out of break

cattrace:
> always

On Windows,There is a breaktime program name "eyefoo", when you are breaking, the program will force the system swtich to it and show it on top (it repeat this action per second), so you cant do anything(it will switch back immediatly), i think the "loop action" is the key @

hovancik:
That's interesting approach! I'll try that

hovancik:
So I have tried to `show()` window when I am in break every few milliseconds but that does not work ;/ 

it should work as I was able to do that with `electron-quick-start` application. No idea what it makes not to work with stretchly, could not find it.

smaheshs:
RSI guard blocks keyboard and mouse inputs once the break starts. Could something similar be done?

MM-John:
I came here to ask for an option the break window to NOT be "always on top". I take my breaks ON the computer, most of the time. They are playful, mentally refreshing, inspirational, motivational breaks, but they require me to click on things and look at windows and type. Example: journaling.

My workaround is to set break window size to 0.5 and it's transparency to 50%. Then I can work behind it. Not perfect, but doable.

EDIT2: another workaround I found is a free, portable utility called [GUIPropView](https://www.nirsoft.net/utils/gui_prop_view.html) by nirsoft (I use dozens of his utilities). With that, I can find the Stretchly window, and turn off "Top Most" and then move around my windows normally. However, at that point the Stretchly window doesnt appear in the Windows 10 Alt+Tab list, nor is there an entry for it in the Windows 10 taskbar. To get the Stretchly window back, I have to use GUIPropView Switch To. Not perfect, but it works.

I'm on Windows 10. v1.4.0.

I'm not sure if I should submit this as its own request.

(Great program, by the way. I wouldn't spend the time to come here and ask if so many features weren't terrific already.)

EDIT: asking for "an option" not to change default behavior

[...]

PR #875, opened 2021-02-28
'actAsRegularWindow' contributor setting implementation

darksworm:
Issue: #437

### Requirements

- [ ]  all new code has tests to ensure against regressions. - :exclamation: **I'm afraid that the only way to test this reasonably is via acceptance tests and as they are all commented out (`test/main.js`) I haven't put in the effort**
- [ ] `npm run test` is error-free. <details><summary>:exclamation: **unrelated tests are failing** </summary> image</details>

### Description of the Change

Added "Make stretchly act as a regular window" setting to the contributor settings. When enabled, you can move, minimize and send-to-back the stretchly break window. You cannot however close the window.

### Test scenario:
1. set "Show breaks in" to "Window"
2. set "Make stretchly act as a regular window" to ON :heavy_check_mark: 
3. "Skip to next" -> "Mini break" via icon in corner of taskbar

#### Results:
-> Window has title bar, break window appears in taskbar, window cannot be closed via ALT-F4 or exit button, window can be minimized, showAsRegularWindow setting appears in contributor settings.

#### Repeat same for "Long break":
-> Same results as previous.

#### Repeat same for "Show breaks in" = "Fullscreen" with both Mini and Long breaks:
->  Window covering whole screen appears, has a title bar, appears in task bar, cannot be closed via ALT-F4 or exit button, window can be minimized.

#### Repeat same for "Long break", same results.

<details>
<summary>Testing on windows - proof</summary>

https://user-images.githubusercontent.com/9987548/109429519-bf20fc80-7a04-11eb-994b-5f6f1bd0d0a0.mp4

https://user-images.githubusercontent.com/9987548/109429523-bfb99300-7a04-11eb-8d34-1e4145bbfc7f.mp4

https://user-images.githubusercontent.com/9987548/109429524-c0522980-7a04-11eb-88da-97f7c10629c6.mp4

https://user-images.githubusercontent.com/9987548/109429525-c0522980-7a04-11eb-9d45-32e55df6b496.mp4

https://user-images.githubusercontent.com/9987548/109429597-2343c080-7a05-11eb-868b-09b05890c3b0.mp4

</details>

<details>
<summary>Testing on linux - proof</summary>

Super+Q is the hotkey to close a window.

https://user-images.githubusercontent.com/9987548/109429994-96e6cd00-7a07-11eb-99ab-b286440def7e.mp4

https://user-images.githubusercontent.com/9987548/109429995-977f6380-7a07-11eb-9549-489a47c2b954.mp4

https://user-images.githubusercontent.com/9987548/109429996-977f6380-7a07-11eb-81bc-81665d54662b.mp4

https://user-images.githubusercontent.com/9987548/109429997-9817fa00-7a07-11eb-9b89-17a8071f643a.mp4

https://user-images.githubusercontent.com/9987548/109430001-9e0ddb00-7a07-11eb-89b9-648ceda47df7.mp4

https://user-images.githubusercontent.com/9987548/109430002-9ea67180-7a07-11eb-8233-e505669dc6a2.mp4

</details>

### Other information

I have no means of testing how this behaves on macos - help with that would be appreciated


darksworm:
I'm having troubles testing this on windows, when I build it for windows with `electron-builder --x64 --win` and launch it, I get these errors:

image

Googling these has not been very helpful - seems to be something to do with having 32 bit depedencies? IDK.

@hovancik have you experienced this issue?

hovancik:
what's your OS, 32 or 64?

darksworm:
> what's your OS, 32 or 64?

64bit

darksworm:
> > what's your OS, 32 or 64?
> 
> 64bit

tried building on windows - that solved the issue.

darksworm:
@hovancik this is ready for review.

README.md
```diff
@@ -173,6 +173,10 @@ To specify the size of the break window, set the value of `breakWindowHeight` an
 #### Check for new version
 If you don't want to ever check for new version, set `checkNewVersion` to `false`.
 
+#### Make Stretchly act as a regular window [![Contributor Preferences](https://img.shields.io/badge/Contributor_Preferences-✔-success)](#contributor-preferences)
+
+If you want Stretchly to act as a regular window (have a titlebar, turn off always on top, be minimizable) set `actAsRegularWindow` to `false`.
```
hovancik: this should be `true`?

app/main.js
```diff
@@ -471,8 +472,11 @@ function startMicrobreak () {
 
     microbreakWinLocal.loadURL(modalPath)
     microbreakWinLocal.setVisibleOnAllWorkspaces(true)
-    microbreakWinLocal.setAlwaysOnTop(true, 'screen-saver')
+    microbreakWinLocal.setAlwaysOnTop(actAsRegularWindow, 'screen-saver')
     if (microbreakWinLocal) {
+      microbreakWinLocal.on('close', (e) => {
```
hovancik, line 490: What is this doing?

app/contributor-preferences.html
```diff
@@ -411,6 +411,11 @@
         <input type="checkbox" class="enabletype" value="isFirstRun" id="showWelcomeWindow">
         <label data-i18next="contributorPreferences.showWelcomeWindow" for="showWelcomeWindow"></label>
       </div>
+      <div>
+        <input type="checkbox" class="enabletype" value="actAsRegularWindow" id="actAsRegularWindow">
+        <label data-i18next="contributorPreferences.actAsRegularWindow" for="actAsRegularWindow"></label>
+      </div>
+      <div></div>
```
hovancik: Let's make changes here: 
image

1. Let's do it similar to this: 
image

For example: 
```
Act as regular window :  When enabled, Break window will become focusable, it won't be 'Always on top' anymore and it will become available in taskbar. 

(checkbox) Show breaks as regular windows
``` 

You are also missing translations. You only need to add English ones. 

You can test this window by changing `isContributor: false` to `isContributor: true` in `main.js`

hovancik (changes requested):

hovancik:
I am trying on Ubuntu 20.04 and when Break is on, the window is not gone when I click on other windows. 

Is that expected? I thought the point was to be able to go to another windows.

app/main.js
darksworm: This prevents the break window from being closed.

README.md
darksworm: yup, nice catch!

darksworm:
> I am trying on Ubuntu 20.04 and when Break is on, the window is not gone when I click on other windows.
> 
> Is that expected? I thought the point was to be able to go to another windows.

no it is not, the boolean value for the second set of `setAlwaysOnTop` calls was inversed. fixed.

app/contributor-preferences.html
darksworm: changed in https://github.com/hovancik/stretchly/pull/875/commits/404d234471252d8e005ec6bae18c706e599efd81

image

app/main.js
darksworm: because:
- allowing the window to be `focusable` lets the users ALT-F4, CTRL-W to close it
- adding the window to the taskbar lets the users close the window by right-clicking it in the taskbar
- adding the titlebar lets the users press the cross on the window to close it

AFAIK there is no way to remove those close buttons in electron besides removing the frame (which I've avoided to let users drag and minimize the window in this mode)

darksworm:
@hovancik this is ready for your review

app/main.js
hovancik: Ah, ok :) Makes sense

app/contributor-preferences.html
hovancik: Looks good! 

Small issue with next block; `Appearance` should be on left side

image

hovancik:
Haven't tested elsewhere, but on my Ubuntu, break is not opened correctly: I do not see the break window, I only see new item in tasks list: 

image

hovancik:
Can you also try quitting the app from tray while break is on? It seems like whole app kinda freezes.

app/contributor-preferences.html
darksworm: Don't know how I missed that!
Fixed in https://github.com/hovancik/stretchly/pull/875/commits/b31b5a70433819498f2ba88fc6f124807d64dfca
image

darksworm:
> Haven't tested elsewhere, but on my Ubuntu, break is not opened correctly: I do not see the break window, I only see new item in tasks list:
> 
> image

Couldn't replicate on windows or DWM on linux, but managed to replicate this on budgie - it was due to the window not being focused upon creation, i guess different window managers handle this differently. 

fixed in https://github.com/hovancik/stretchly/pull/875/commits/3f7d5ec8b05b0ee14a2af68ab0b5d29a1bff975d (also noticed the macos kiosk mode settings, adapted those as well to work in line with the new setting)

darksworm:
> Can you also try quitting the app from tray while break is on? It seems like whole app kinda freezes.

Yup! Turns out the code which stopped the break windows from being closed also stopped the whole application from being closed while they were visible.

I've changed it so that when calling `app.quit`, the break windows will close.

---
I've addressed all issues, waiting for further comments.

app/main.js
```diff
@@ -450,10 +455,15 @@ function startMicrobreak () {
...
     microbreakWinLocal.once('ready-to-show', () => {
-      microbreakWinLocal.showInactive()
+      if (showBreaksAsRegularWindows) {
+        microbreakWinLocal.show()
+      } else {
+        microbreakWinLocal.showInactive()
+      }
+
       log.info(`Stretchly: showing window ${localDisplayId + 1} of ${numberOfDisplays()}`)
       if (process.platform === 'darwin') {
```
hovancik, line 465: Did you test on macOS, then? Elsewhere you wrote you can't test there

darksworm: No, I did not. Sadly, I don't have access to a macOS device.

hovancik: Fullscreen does not seems to work, will try investigate later

darksworm:
I've rewritten the end-to-end tests (`main.js`) to make them work and added partial coverage for this feature.

I cannot however get the tests to run on windows and ubuntu in the github action, so i've changed the CI to launch the end-to-end tests only on macOS.

Although not great, it does seem better than not running them at all. They won't be included in the coverage, but I feel that's OK - otherwise e2e tests generate fake coverage anyway.

app/main.js
hovancik: I think it should be 
```
      if (process.platform === 'darwin') {
        if (showBreaksAsRegularWindows) {
          microbreakWinLocal.setFullScreen(settings.get('fullscreen'))          
        } else {
          microbreakWinLocal.setKiosk(settings.get('fullscreen'))
        }
      }
```

hovancik:
Thanks for the tests! 
What was the error? I believe I disabled the tests because of this: https://github.com/electron-userland/spectron/issues/656

darksworm:
> Thanks for the tests!
> What was the error? I believe I disabled the tests because of this: [electron-userland/spectron#656](https://github.com/electron-userland/spectron/issues/656)

Yeah, i encountered that error - it seems that the "promiseness" transfer just doesn't work - I rewrote the tests to just use `await` instead.

---

The e2e tests fail on github actions with this currently:
```
unknown error: Chrome failed to start: exited abnormally.
  (unknown error: DevToolsActivePort file doesn't exist)
  (The process started from chrome location /home/runner/work/stretchly/stretchly/node_modules/spectron/lib/launcher.js is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
      at Object.startWebDriverSession (node_modules/webdriver/build/utils.js:34:15)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

Most articles suggest adding some chrome driver options: `--headless,--no-sandbox,--disable-gpu,--disable-dev-shm-usage` but these make no difference.

Also tried setting up a virtual display with `xvfb` - didn't help either.

The tests run just fine on my local linux machine - that makes me believe this has to be be something to do with the headless github actions environment.

app/main.js
darksworm: Yup! That makes a lot of sense to me - I've applied those changes here - https://github.com/hovancik/stretchly/pull/875/commits/a66893e72dc6d7115bfebb52f0ab4b4f8a5e4db0

hovancik (approved):
I guess we are ready to merge, then?

app/main.js
```diff
@@ -471,8 +485,13 @@ function startMicrobreak () {
 
     microbreakWinLocal.loadURL(modalPath)
     microbreakWinLocal.setVisibleOnAllWorkspaces(true)
-    microbreakWinLocal.setAlwaysOnTop(true, 'screen-saver')
+    microbreakWinLocal.setAlwaysOnTop(!showBreaksAsRegularWindows, 'screen-saver')
     if (microbreakWinLocal) {
+      microbreakWinLocal.on('close', (e) => {
```
hovancik, line 490: So looks like this is causing issues on macOS: #903

Issue #717, opened 2020-10-01
Time for next break in icon


hovancik:
I would like to investigate what's mentioned in this comment:  https://github.com/hovancik/stretchly/issues/421#issuecomment-507147712

<img width="138" alt="image" src="https://user-images.githubusercontent.com/4745859/118631842-2cf0ef00-b7d0-11eb-8d3b-c2911c3c7b40.png">

In the past I closed few issues that were asking to add text to the icon, doing the same as I believe icon is too small to communicate that. 

Approach mentioned in comment might have the same issue, though.   

I'll contact designer once I get back to this.

PR #957, opened 2021-05-11
Minutes remaining in tray

senpl:
This should solve https://github.com/hovancik/stretchly/issues/717 for most users.

This works for me. But to decide how it should be finally implemented.
My Proposition is as follows, yellow numbers.
Pictures from 0 to 9 in folder /app-icons/numbers
They could get generated into /app-icons/numbers from 0 to 99.
When option is selected numbers are shown in try icon.
It only shows time to big break as to me it seems like most important.
 

### Requirements

- [x] issue was opened to discuss proposed changes before starting implementation. It is important do discuss changes before implementing them (Why should we add it? How should it work? How should it look? Where will it be? ...).

### Description of the Change

It is now possible to show times left in tray icon. I added option for long break and it is by default turn off.
Folder with basic pictures is added. Their number is from 0.png to 9.png . From them it could be possible to generate all other from 00.png to 99.png . This pictures in generated folder are used to show correct minutes left to long break. Currently method for generation of number is not on, but it could be used in final version

### Verification Process
- How did you verify that all new functionality works as expected?
It works on my machine. 
- How did you verify that all changed functionality works as expected?
Changes are almost fully in my class so not much is affected.
- How did you verify that the change has not introduced any regressions?
Not needed.

I only test it in Arch linux, I do  not have mac to check it.
I use it daily so during work . It also work with pause, but refresh is not instant.

### Other information

There are many things to decide.
1. If images color yellow is acceptable?
2. In which place of options this option should be placed?
3. Is users should be able to generate their own image numbers from app?
4. How many generated pictures to store and how to update them?(do generation of image during work is optimal? or should we use fully generated images)?


senpl:
To make it even more lightweight images with numbers and icons could be generated, It would be more complicated, then current implementation. But would not generate pictures during program run(which seems nice). The brutal force solution would be to generate all numbers with all possible icons. Or add some additional generation when icon is changed. Still current implementation works and is not terribly optimized.

hovancik:
Hi, sorry for late reply, busy :) 
Haven't had chance to run your code, but I would like to have it similar to what other posted from some app (on Stretchly logo): 

image

Any chance you have skills to do that? 

Also to consider, many people do not use Long Breaks, so I think it should be for both.

senpl:
Sure, it could be done. But I would like to options with numbers as well. But as many logo options exist I guess adding this over existing logo would make it less visible. And I will definitly not test it on mac.
I think if this will work for long breaks then adding it to shorter breaks could be next step (after all most people are interested in long break, not in their short break).
I would also like anwser to my 4 questions.

senpl:
And I propose maximum circular break minutes to be 60 like in clock minutes, bigger will be just shown as 60.
To short breaks I have no idea how to scale them so this would be done in second step(any proposition are welcomed). First just for minutes.

hovancik:
Hey sorry, I won't have time this weekend, sorry for making you wait.

senpl:
Ok. I am close to finsh round circle for long breaks.
My main questions:
-Should time of circle should be based on user break time or real minutes.
(so if users set 40 minut does full circle be shown after 40 minut ?
-How smal intervals are really needed. Currently it is every 15 minutes on clock, so not that much. It has to be somehow limited coz the more frequent updates the more pictures has to be prepared, which is at this part not really needed. In final version number of intervals could be increased, I just not want to make final version without general approval of design.

senpl:
And one more question, how do I get time that is currently set for long break every minutes?

senpl:
> And one more question, how do I get time that is currently set for long break every minutes?

Ok. I figure it out its (settings.get('breakInterval') + 1 ) * 10 . Not really nice to store that important value. Maybe this should be refactored to minutes that could be set other then rounded of 10?

senpl:
Ok. currently presented round icon is too small to be implemented as on picture. I could increase it without rounding gray circle and one pixel of boarder, it will not looks like on picture, but is the best that could be visible on small tray icon.

hovancik:
Hi @senpl , sorry for taking my time, life just bit crazy for me right now ;/ 

So how Stretchly works, to explain how Long Break gets calculated: 
- it is always N number of Mini Breaks followed by one Long Break
- user can specify interval and length of Mini Breaks
- user can specify N number and length of Long Breaks



In general, as for MVP, I would keep it as option available only in configuration file. 

For icons: 
- I think there are 2 sizes for icons:  one for Mac and second for Linux/Windows
- there is also Colorful, monochrome and inverted monochrome version (mac can autoinvert monochrome, so that is not needed for Mac, only for Win and Linux)
- would be nice to have them as SVGs as I've done with the current icons so that they can be changed (https://github.com/hovancik/stretchly/tree/trunk/graphics)

I guess to increase visibility, we could not show Stetchly logo? 
Or second option is not to show both Mini and Long breaks, but only the next one. We reuse same code and images for both. Could be configurable whether to show both mini and longs or only for long break as is your usecase. Makes the most sense to me now. 

I think values should be relative, so that if next break is in 40 minutes and interval is every hour, we are at 40/60 position in circle (I think it makes sense as I see many users who have let's say only Mini Breaks every 2 hours, so much bigger intervals than 60 minutes). Let's not do many images as you say, we can decide on update interval later. 

With generated/saved images: I am not really sure. Doing them with code would save the app size, but how much resources does it take? We would need to try.

senpl:
Ok. Now round timer works from minutes left to total minutes of long break ratio.
Still not done is translation of timer select and it not shows what is currently selected. But I do not know how to make it, so I left it not translated. All other thigs should work. If somethings should be changed let me know.

senpl:
This version of logic should now work as expected.
Now things to consider:
1. Background circle of round icon is not really needed from my point of view, or their transparency level should be really low, current makes logo just washout of colours.
2. Width of round circle line probably should get more width to be better visible, still the bigger long breaks circle, the smaller short time circle so this is always tradeoff. As this version is long breaks only this might be good make it more visible.
3. Position in options. this could be easly changed, still this has to be somewhere.

hovancik:
I've pushed the conflict merge so I can run it locally, will be looking into your comments and code.

hovancik:
Ah, so no, I can't seems to push to your branch.

app/main.js
```diff
@@ -933,9 +934,18 @@ function createPreferencesWindow () {
 }
 
 function updateTray () {
-  updateToolTip()
-  appIcon.setImage(trayIconPath())
-  appIcon.setContextMenu(getTrayMenu())
+  updateToolTip();
```
hovancik: I think that this should stay as it was before, so those 3 calls: 
```
updateToolTip()
appIcon.setImage(trayIconPath())
```
[appIcon class](https://github.com/hovancik/stretchly/blob/trunk/app/utils/appIcon.js) should stay the only one that is responsible for the icon in tray. We can then expand on [tests](https://github.com/hovancik/stretchly/blob/trunk/test/appIcon.js) as well.

`AppIcon#trayIconFileName` should return correct path to the icon. I would need more time with your code, but it seems to me now that we need to  extend  `constructor ({ platform, paused, monochrome, inverted, darkMode })` with 2 more values, something like: 
- `remainingModeString` - which would be `""`, `"Circle"` or `"Number"`
- `remainingTimeString` - which would be `""`, `"1"`, `"2"` etc...

Then first code will become the second: 

```
      if (this.platform === 'darwin') {
        return `trayMacMonochrome${pausedString}Template.png`
      } else {
        return `trayMonochrome${invertedMonochromeString}${pausedString}.png`
      }
```
-> 

```
      if (this.platform === 'darwin') {
        return `trayMacMonochrome${pausedString}${remainingModeString}${remainingTimeString}Template.png`
      } else {
        return `trayMonochrome${invertedMonochromeString}${pausedString}${remainingModeString}${remainingTimeString}.png`
      }
``` 

`remainingModeString` and  `remainingTimeString` could get calculated by some class based on current state and preferences and return values described up. That is then passed to constructor of `AppIcon`. 

 To go with my example code, let's say we have 2 minutes to break, we are on Mac, monochrome icon mode is on and we wanna show numbers; the resulting file would be: 

> return `trayMacMonochrome${pausedString}${remainingModeString}${remainingTimeString}Template.png`

`trayMacMonochromeNumber2Template.png`

Let's say breaks are paused, we are on Linux, and inverted monochrome, we wanna show circles. All is good and we get correct values as `${remainingModeString}${remainingTimeString}` are both empty strings. 

> `trayMonochrome${invertedMonochromeString}${pausedString}${remainingModeString}${remainingTimeString}.png`

`trayMonochromeInvertedPaused.png`

As mentioned earlier, this can be nicely tested.

senpl:
@hovancik I updated repo access and now edit should be possible.

hovancik:
>     Background circle of round icon is not really needed from my point of view, or their transparency level should be really low, current makes logo just washout of colours.

I think for both Circle and Number versions, we should use the same colors as Paused icon is implemented. There is good contrast, I think: (they were done by professional designer)

image

>    Position in options. this could be easly changed, still this has to be somewhere.

Let's talk about this later, it really depends how much gets implemented and how sure we are it will stay that way. For minimal version, it's easier to keep in config file only.

senpl:
> Number versions, we should use the same colors as Paused icon is implemented.

Black on black for numbers is not really well visible. So yellow is better, but i could provide color original and yellow, but that would be just another options and assets. Or more advanced color choice for users that would overrides assets, but this seems like overeengineering. Color for circle i do not care but one color is simpler for anyone. But in final version i could provide 3 circle version, but This could also be done by user who need it.

hovancik:
There are multiple versions of icons, as needed for different OS. Additional to the original SVGs (https://github.com/hovancik/stretchly/tree/trunk/graphics), here are the generated one from that: https://github.com/hovancik/stretchly/tree/trunk/app/images/app-icons  

## 1. Monochrome 
This is for Mac, and it needs to end with Template.png so that macOS knows it can change it between light and dark. 
For Linux and Windows this is default when monochrome is selected. 
image

## 2. Inverted monochrome
This is for Linux and Windows, when Inverted Monochrome is selected, as they do not do it automagically as MacOS. 
image

## 3. Regular Light
All of the OSes, when we are in OS Light mode
image

## 4. Regular Dark
as above, for Dark mode 

image

## In general
- macOS icon size is 16x16px 
- macOS also needs @2x.png (32x32px)
- Linux and Windows need 32x32px icon size

=> to keep it consistent we should use the same colors as above, regular icon <=> paused icon

senpl:
Why not allow users to choose whatever color they want ??? This seems simpler then keep 3 colors version just for conistency with design not even created for this task in mind ??

hovancik:
> Why not allow users to choose whatever color they want ??? This seems simpler then keep 3 colors version just for conistency with design not even created for this task in mind ??

Not sure what you mean. 

User is picking whether they want monochrome or colour tray icon. Each has then version for Dark/Light mode. So we need the same for this feature. 

Fot the color of numbers/circle: 

What we have here is basically `icon` and then some `change` to icon. Previously, there has been regular icon as `icon` and then paused icon when we added paused sign as the `change`. With your code, we are adding 2 more `change`s : numbers and circle. It makes sense to me too keep the color of `change` the same across all features for consistency. 

If you can make it configurable, then I am all in, but it seems to me like a hard work, considering what different sizes and everything else is required. With one exception, of course, using colors for Monochrome would be against its point: 

image

senpl:
Ok. Now I get that
1.  Monochrome should keep one color, but for visibility purpose it is not possible to make them in same color as icon(they just will not be visible) . So I could make it visible with use of some degree of gray.
2. Regular Light second number when will be black will not be visible. So I stay on yellow option with instruction how to change it if user need it.
3. Regular Dark and Inverted monochrome ok. This could be white, this will not be as well visible as yellow but still at least visible.

hovancik:
Kinda. I think designer did it really well. My pictures above might not look so good, because they are against the dark background.

For example colorful version for dark theme: 

image

Then the paused version for the same:

image

I think if you were to use the same color for numbers/circles as it is used for pause sign, it would be very well visible and we would have consistency. 

The colors of default icon seems a bit different between normal and paused state, when you carefully. But from what I can see in source, it's because there is added transpaecy in paused state: 

image

image

Colors are exactly the same. 

Also the one you say looks bad, seems fine to me (it's for light theme, so light background):

image

image

Same story as with dark one, colors are the same but there seems to be transparency added: 
image

I think others are also fine once you put them into correct context: 

image
image

So I think we should do the same:
- have SVGs for icons that we can then transform to all those required sizes (https://github.com/hovancik/stretchly/pull/957#issuecomment-856699322) - I guess we could have some tool/code to generate those, as it's many many many 
- they should be the same as the paused state, so with that transparency or whatever that is

I think size is not issue, currently I have a lot generated stuff and it's super small: 
image

Fonts are much bigger issue, as there are many exotic languages :D 
image

I could try asking the designer that designed those icons to create the new ones or try to find some other designer.

senpl:
This is newst version. It should work when all pictures are in folders. Sure now some temp files could be cleaned. 
Code is simpler, but much more pictures are generated. This will be problematic only during icon change so not frequently. I leave generation of pictures method but they are not used. Only name taking is in usage now.
I reworked parametes as you suggested. Still pictures for keeping it clean are in respective folders.
All path return could now be moved to appIcon, but I guess it is not really needed.
Let me know if this is close to what is expected to be accepted.

hovancik:
Thanks, will take a look. I've updated your branch with latest `trunk`

hovancik:
Haven't had chance to check the code, yet, but running with numbers, I noticed it seems to override all other icons? 
- when break is on, icon is "0" icon
- when my system goes to Do Not Disturb, I am also seeing "0" icon
- when I pause Indef, i also see "0" icon
- when I pause ie for 5 hours, I see "3" icon

I would expect this would coexist with other icons, so 
- when break is on, I am back to normal icon
- when breaks are paused (dnd, natural, manual...), I am back to paused icon

I also think we should never see "0" icon: because of JS, ie in Break sometimes values can be negative, but Stretchly never goes to less then "1 second remaining"

senpl:
I guess I fixed paused mode. DND mode not worked on my machine so I could not reproduce it, but in theory it should be fixed as well.

app/main.js
```diff
@@ -39,6 +41,7 @@ let microbreakIdeas
 let breakIdeas
 let breakPlanner
 let appIcon = null
+let appIcon2 = null;
```
hovancik: This won't be needed anymore. 

Also, I am using [standardjs](http://standardjs.com/) but I guess your editor is making some auto-changes that go against standardjs. You can see errors with `npm run lint` but I think we can leave this as a final step.

hovancik:
I am on Linux, with dark theme, numbers version, default preferences and I noticed that when I am on icon "10", all is fine (10 is in light color), but then following icons have number in dark color.

senpl:
I fixed numbers for dark color 0-9 default tray. I updated lints and I added parameters to tests. Hope all now works. Let me know what else should be fixed.

hovancik:
Monochrome on Mac is also not working, it needs a special name, see here: https://github.com/hovancik/stretchly/pull/957#issuecomment-856699322

app/main.js
```diff
@@ -371,7 +393,10 @@ function windowIconPath () {
     monochrome: settings.get('useMonochromeTrayIcon'),
     inverted: settings.get('useMonochromeInvertedTrayIcon'),
     darkMode: nativeTheme.shouldUseDarkColors,
-    platform: process.platform
+    platform: process.platform,
```
hovancik: Could we not send real data here, but some empty value? I see `windowIconFileName ()` only uses `inverted`, `darkMode` and `monochrome`. Just to see from `main` that it is not used.

app/main.js
```diff
@@ -930,9 +979,12 @@ function createPreferencesWindow () {
   preferencesWin.on('closed', () => {
     preferencesWin = null
   })
+  setTimeout(() => {
+    preferencesWin.center()
+  }, 0)
 }
 
-function updateTray () {
+async function updateTray () {
```
hovancik: Just to know for myself: why was this needed?

app/utils/trayWithText.js
```diff
@@ -0,0 +1,245 @@
...
+class TrayWithText {
+  iconWithNumber (
+    imagePath,
+    minutesToLongBreak,
+    darkModeString,
+    monochrome,
+    platform,
+    invertedMonochromeString
+  ) {
+    const minutesOnTray = minutesToLongBreak
```
hovancik: What behaviour we should see when we go over 99? 

Right now I think we have 49 when we have 149 and such. I believe better would be not to show anything in such cases.

senpl: I guess minutesToLongBreak are always below 99 as I use modulo in returned. I could show 99. Not showing anything is not really what user would expected(or it would require additional description). Anyway I could easly make it 99 for numbers bigger then 99.

app/main.js
senpl: In previous version I used image generation and generations could take time. I do not wanted to return path that was not ready coz it would not work. Anyway in current implementation it is not needed and could only be needed when short break would be implemented not by brute force approach.

app/main.js
senpl: Sure. I followed what was before, so pause, platform and all below could empty values. To what value set them ???

senpl:
> * macOS also needs @2x.png (32x32px)

I not found usage of 2x.png in code. And I do not have macOs. So I have no clue how to and what for to use it. 


> Monochrome on Mac is also not working, it needs a special name, see here: [#957 (comment)](https://github.com/hovancik/stretchly/pull/957#issuecomment-856699322)

In my version for Mac textWithTry return:
`if (platform === 'darwin') {
        return `${imagePath}traytMacMonochrome${minutesOnTray}.png`
      } `
As long as this files are in folder /generated-numbers it should work. Could I precisely know what not works, and in which mode ???

app/main.js
hovancik: I think `null` should be fine?

app/main.js
hovancik: Ok, let's put it back to what it was then. I will forget in time and will be afraid to change it :)

app/utils/trayWithText.js
hovancik: Well, seems to that showing nothing is "less wrong" then showing wrong number. 193 and 93 is a big difference. But ok, let's keep it...

hovancik: I take that back, I am testing with 105 minutes and it is showing 5. Let's make it at least 99. But if nothing is easier, let's show nothing

hovancik:
`@2x` is picked automatically by OS with better resolution screens, it is not in code.

I mean  "and it needs to end with Template.png" as you can see in `app-icons` folder. It's for monochrome version, mac can change between regular and inverted automagically if named correctly.

senpl:
ok. so @2x could be just copy of existing normal icon with Template names for monochrome. I could copy them or just use mac for every other. Second option will take less space but will not be as clear so I guess another copy of pictures is a way to go.

senpl:
Ok. i added all x2 pictures even if I am not sure if their needed.

app/utils/trayWithText.js
senpl: I prefer 99 coz user who turn this feature see that it works. empty solution would be less intuitive in my opinion. I just pushed version changes that should make it 99.

hovancik:
Does `npm run test` works  for you? CI is failing and I am seeing the same locally on Linux. When changing to my `trunk` test passes.

senpl:
I guess it's only default setting comparison. It works when "Number" Is used. From some reason compare of empty strings not gives good results. So I changed default, but of course probably string like '-' or ' ' should works.

app/utils/defaultSettings.js
```diff
@@ -57,5 +57,6 @@ module.exports = {
   pauseForSuspendOrLock: true,
   pauseBreaksShortcut: '',
   resumeBreaksShortcut: '',
-  screen: 'primary'
+  screen: 'primary',
+  breakIconType: 'Number'
```
hovancik: Can we use some better name? This name suggest "What icon type should we use in Break?".  

To keep it consistent with the rest of the app, I think UI should be similar to how Sounds work: 

image

> Time to break: (checkbox) Show remaining time in menubar icon
(select) Minutes (select) Circle countdown

Therefore probably 2 preferences are needed: something like `showTimeToBreakInTrayIcon`(default `false` so users don't get surprised) and `timeToBreakTrayIconType` (`number`, `circle`). 

This should probably be under "Theme", not "Schedule", but I think until we have better icons for the circles, we should only keep it under "Advanced preferences". Could be added to "Contributor preferences" meanwhile

app/utils/trayWithText.js
```diff
@@ -0,0 +1,262 @@
...
+    let minutesOnTray = minutesToLongBreak
+    let minutesAsNumber = 99
+    try {
+      minutesAsNumber = Number(minutesOnTray)
+    } catch (er) {
+      minutesAsNumber = 99
+    }
+    if (minutesAsNumber >= 99) {
+      minutesOnTray = '99'
+    }
```
hovancik: This class still looks like complication to me. I think what I suggested here is better: https://github.com/hovancik/stretchly/pull/957#discussion_r645984672 

`TrayWithText` should be only responsible for generating images (therefore probably renamed to something like "time in tray icon generator" or whatever) and `appIcon` being central place where we decide what icon is currently in tray. I don't see any advantage in making decision in 2 separate places. Code and contructor values seems duplicated.

app/utils/trayWithText.js
```diff
@@ -0,0 +1,262 @@
...
+    } else {
+      if (platform === 'darwin') {
+        return `${imagePath}traytMac${darkModeString}${pictureNo}.png`
+      } else {
+        return `${imagePath}trayt${darkModeString}${pictureNo}.png`
+      }
+    }
+  }
+
+  async generateNumbers (color = 'b', mac = 'm') {
```
hovancik: There is no test coverage for the icon generation.

app/utils/trayWithText.js
```diff
@@ -0,0 +1,262 @@
...
+        log.debug(e + 'safely ignored error')
+      }
+    }
+  }
+
+  async generateCirclesWithTray (
+    color = 'b',
+    createdName = 'traytMacMonochrome',
+    mac = 'm',
+    iconPath = '/home/m/p/stretchly/app/images/app-icons/traytMacMonochrome.png'
```
hovancik: This should be relative path, right?

app/utils/utils.js
```diff
@@ -69,6 +69,12 @@ function formatKeyboardShortcut (keyboardShortcut) {
   return keyboardShortcut.replace('Or', '/').replace('+', ' + ')
 }
 
+const minutesRemaining = function (milliseconds) {
```
hovancik: There is no test coverage for this.

package.json
```diff
@@ -123,17 +115,31 @@
...
     "electron-log": "^4.3.5",
     "electron-store": "^8.0.0",
-    "i18next": "^20.2.2",
-    "i18next-node-fs-backend": "^2.1.3",
-    "luxon": "^1.26.0",
+    "i18next": "^20.3.1",
+    "i18next-fs-backend": "^1.1.1",
+    "luxon": "^1.27.0",
     "meeussunmoon": "^3.0.3",
+    "merge-img-vwv": "*",
```
hovancik: Is this needed in production build?

test/appIcon.js
```diff
@@ -22,7 +23,8 @@ describe('appIcon', function () {
       monochrome: false,
       inverted: false,
       darkMode: true,
-      platform: 'linux'
+      platform: 'linux',
+      remainingModeString: ''
```
hovancik: There should be test that will make sure under different remaining minutes we get the correct icons for each platform, similar as it is done with dark mode etc. 

There should also be test that when break is on, the icon is not "1", or "0" but the regular one. In fact, we should never see icon "0".

hovancik (changes requested):

hovancik:
Test still seems to fail :(

app/utils/trayWithText.js
senpl: 1. no. Images always have to have real path before operations with them. But it could probably be build with use of path lib. As this is methods for generation only and it is default parameter it could be like that, but of course to work for everyone it should be combination of path+ 'app/... '

app/utils/trayWithText.js
senpl: And image generations are not really used in app. Once generated it's no longer needed. They could only be needed when icon changes for some reason. So I guess it's not needed.

app/utils/trayWithText.js
senpl: Now it could be moved to appIcon. Your old suggestion would make one big folder with imageIcons that would make it hard to know what is where. Currently all images are separated by functionality and it's still sometimes bit to much files in folder. I could copy them  all to one folder but it would not make it clearer. You would get all in one folder, but no. of images will be hard to find anything. So I strongly recommend let current folder structure. In current implementation there is no generation at runtime, so as long as there are files in correct places all works.

app/utils/defaultSettings.js
senpl: I do not have contributor preferences AFAIK. I try to rewrite them to checkbox with select and place them in "Theme".

package.json
senpl: In current implementation no. But if you want to combine short breaks circle then it would be useful.

test/appIcon.js
senpl: Done. Cases with pause I not check coz they in first if. But if their really needed they could be done.
Also move all logic to AppIcon can be done easly.

app/utils/utils.js
senpl: Done.

app/utils/defaultSettings.js
senpl: Done. In theme. But could be now moved to any place.

senpl:
I investigated build error and it's not my error. It's from previous merge request. Anyway adding  --timeout 10000 to test fix issue so this means timeout for some tests are just to limiting(strangly on my fast machine it usually gets into 2000 time limit, but on CI machine it probably need bigger timeout). Sure timeouts could be fixed for every test to skip default timeout. But I guess increasing test timeouts not hurt anyone until tests are written better. If you need more infomration let me know.

senpl:
Ok. I added longer timeout only to 3 tests that got problems. Now not all test should require longer timeout. still UntilMorning test sometimes failes so it's flake test and should not be used in CI.

test/appIcon.js
hovancik: Do you wanna try to move it? I think it would be better

package.json
hovancik: Ok, can you then remove the unused stuff? With new things it's hard for me to be sure :)

app/utils/trayWithText.js
hovancik: If this is used then I would prefer to have `path` so that we don't need to change it when running it.

app/utils/trayWithText.js
hovancik: Ok, I guess that's fine. I will probably try getting designer  to make something in the future

app/utils/trayWithText.js
hovancik: I don't mind many pics under one folder as I rarely touch them and I expect similar for the future. There is much higher chance that I will be changing `AppIcon` and stuff around icons so better for me to have one class to manage everything

senpl: I moved all to one class. Still I not moved pictures. Currently in app/images/app-icons is 27 pictures. If I moved all new one there will be 27 + 1000 + 92. Are you 100% sure you want them all in this one folder? This will not be big rework to move them, it will just make finding basic images much harder to find.

app/utils/trayWithText.js
senpl: They are not used and should not be in test coverage. I could safely remove them, I left them just for conviniece of generating new icons version if in future will be needed.

package.json
senpl: removed.

test/appIcon.js
senpl: Done.

senpl:
I would like to left TextWithTray as generators out of test coverage. As their not used and could be used only by developers. I could move them to separate folders like generators. Or their could be moved to separate project like Stretchly helpers. Also let me know if I should finally move all pictures to one folder.

app/utils/trayWithText.js
senpl: I moved it all in but this could be of course reverted.
e0830cb

senpl:
Any progress on that. Anything else to change?

hovancik:
Hi @senpl , I was changing job so my time and energy were below 0 last month ;/ I should have more time now

app/utils/trayWithText.js
senpl: I guess all is as expected, if not, let me know what change.

hovancik:
In the end, I decided to simplify the code a lot, because I was struggling with getting through understanding everything. For now, I've put it into advanced settings and removed the circle icon. In future, I would like to get some designer to make the icons and them move it into regular preferences. I talked with my original designer, and he was against this feature altogether, as he didn't like how icons would look like and suggested just have icon with number (Linux, Windows) and on macOS show time next to the icon (which is possible). I will have to think more what to do with this.

Issue #984, opened 2021-06-26
[Feature request]: Add systemd user unit so application can auto start on linux


unquote7083:
### Version

- [X] I'm using version 1.7.0

### Existing requests

- [X] I've checked Existing requests

### What operating system are you using?

Linux

### Operating System Version

all systemd-based

### Problem description

No systemd unit means users will need to create their own, or use other means, or have no auto start functionality on linux.

### Possible solution

Add systemd user unit so application can auto start on linux

### Alternative solutions

XDG Autostart might be an alternative

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

hovancik:
Hi, I am using `electron-builder` to make builds. Let's try to find if it can somehow do it. 

In general, I am using https://www.electronjs.org/docs/api/app#appgetloginitemsettingsoptions-macos-windows and there is also an item in Preferences, but this supports only Windows and Mac so it is hidden in Linux. 

I would prefer to have solution that support more general linux approaches, not only systemd. There might be some library that does that. So that we can then also enable "Start at login" option for Linux.

hovancik:
This seems to be maintained but would be good to spend more time looking at different options: https://github.com/Izurii/easy-auto-launch

unquote7083:
Thanks. I think either way it would be beneficial to have the systemd unit file included, as it does nothing on its own, doesn't need any sort of regular updates, and is a very small file.

nihil-admirari:
`systemd` unit starts regardless of whether GUI is available or not.

> XDG Autostart might be an alternative

Starts only when desktop environment is running.

hovancik:
closed in #990 

won't get distributed in app, as I could not find how it works with electron-builder, but people can check it in repo

PR #985, opened 2021-06-26
Add systemd user unit for linux builds

unquote7083:
Issue: #984 

### Requirements
- [X]  issue was opened to discuss proposed changes before starting implementation. It is important do discuss changes before implementing them

### Description of the Change
Add systemd user unit for linux builds. This doesn't do anything on it's own, the user or package manager can choose to do something with it, which will allow for auto-starting stretchly on system login or whenever the user chooses to start it by using `systemctl` tool.

### Verification Process
N/A - no new functionality added

### Other information
Not sure the best place to put the file, so I make a new distribution directory.

### DIFF
--- a/distribution/stretchly.service
+++ b/distribution/stretchly.service
@@ -1,0 +1,14 @@
+[Unit]
+Description="Start stretchly"
+After=default.target
+
+[Service]
+Type=simple
+ExecStartPre=/bin/sleep 1
+ExecStart=/usr/bin/stretchly
+Restart=on-failure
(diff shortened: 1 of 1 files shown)


hovancik:
I would rename `distribution` to `examples`. Only `app`gets distributed but not really sure how that works...

hovancik:
I've merged this in #990

unquote7083:
`examples` is probably more appropriate anyways, thanks!

PR #1023, opened 2021-09-08
Add a 30 minute pause option

### LINKED ISSUE #964: [Feature request]: Add 30mins option to "Pause Breaks"
### Version

- [X] I'm using version 1.7.0

### Existing requests

Windows

### Operating System Version

Edition	Windows 10 Pro Version	20H2 Installed on	‎16/‎1/‎2021 OS build	19042.985 Experience	Windows Feature Experience Pack 120.2212.2020.0

### Problem description

Need to pause breaks for a short period of time to quickly finish a near ending task. However, min pausing of the break is 1 hour which is too long as users would forget to take their break and then began the next task there-by repeating the cycle endlessly - rendering the application as irritating/useless.

- Skipping to the next break forces the user to take a break now when the user intends to finish up the near ending task before taking a break
- Forcing the user to endure (break notifications) which can be distracting to the ending task & then forcing the user to click postpone break would break the concentration of the user.

### Possible solution

Add 30mins option to "Pause Break"

### Alternative solutions

- Allow users to specify custom values in "Pause Breaks"
- Allow users to specify custom values for postponement of break (temporary solution - might affect other intended purposes of application)
- Set Variable schedules (very hard to implement & does not meet other Edge cases)

### Additional information

I understand the purpose of the application and how it forces users to take breaks and I intend to build into a routine with it. However, being a novice/new user to the application, I find it very irritating and counterproductive. I usually end up either:
- positioning break for the max period (and taking my real breaks out of sync with the application),
- completely closing the application (and forgetting that I need breaks/its deactivated), 
- using the 1 hour break option (in which case I usually start my next task and get stuck in a cycle of postponing breaks)
- Allowing the breaks to pop-up and click postpone/skip break until I lose track of the breaks I skipped or get irritated with it and select a different option

(Break setting on original 20/60)
I understand the possibility of lengthening the break intervals, however, resist it as aiming to fit into it (as recommended officially as well). Also, the above situation happens during deep concentration works that are near ending only (however, given the high occurrence rate (given application always on), it causes difficulties utilizing/integrating the application.)

### Code of Conduct

jwood13:
Issue: closes #964

### Requirements

- [ ]  all new code has tests to ensure against regressions. (Nothing major to test)
- [ ] `npm run test` is error-free. (Until morning tests failing, but were already failing)

### Description of the Change
Added a 30 minute option to the toolbar pause breaks menu

### Verification Process
Clicked on the added option in the menu, and observed the tooltip switch to 'Paused - Resuming in about 30 minutes'

Checked the language on the option in en and de languages.

### Other information
Used the utils for localisations rather than add a new term

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - new break ideas
 - advanced option for showing minutes remaining in app tray
+- 30 minute pause option
 
 ### Changed
 - improve tray status messages
--- a/README.md
+++ b/README.md
@@ -407,6 +407,7 @@ You can help to translate Stretchly on [Weblate](https://hosted.weblate.org/enga
 - Abhilash Mandaliya, [@abhilashmandaliya](https://github.com/abhilashmandaliya)
 - Masi, [@The-Coding-Classroom](https://github.com/The-Coding-Classroom)
 - Saksham Sharma, [@ssaksham](https://github.com/ssaksham)
+- Jared Wood, [@jwood13](https://github.com/jwood13)
 
 Also see Github's list of [contributors](https://github.com/hovancik/stretchly/graphs/contributors).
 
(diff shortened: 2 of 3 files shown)


hovancik (approved):
Thanks!

PR #1092, opened 2022-01-24
Allow the next break or microbreak to be scheduled from the CLI

### LINKED ISSUE #1091: [Feature request]: Allow the next break or microbreak to be scheduled from the CLI
### Version

- [X] I'm using version 1.9.0

### Existing requests

Linux

### Operating System Version

Arch Linux

### Problem description

I use `crontab` and the `stretchly` CLI to align Stretchly's breaks with my daily schedule (and to remind me to use them in the ways I intended at the start of the day 😉 😬 ...)

Stretchly's CLI implementation has come a long way (kudos!), but today I found myself unable to modify the timing of the next break, which means anytime I reboot / switch user / change device (pretty regular occurrences for me), I lose the benefit of whatever my next reminder would have been if I'd stayed logged in.

It's easy enough for a script to calculate the minutes remaining until the next block of time starts, and I can pass this to the `pause` command, but Stretchly doesn't start a break when it resumes, so that doesn't help.

I could add cron jobs to activate each break, but that would limit the usefulness of Stretchly's tray menu countdown and introduce other side-effects (including a possible race condition), so solving it another way would be preferable.

### Possible solution

Is the Stretchly community amenable to the `mini` and `long` CLI commands supporting an optional time interval to wait before skipping to the given break? Something similar to the `pause` command's `--duration` option, but without `indefinitely` or `until-morning`?

I'll open a PR with my implementation of a `--wait, -w` option as a starting point (written quickly to meet a personal need, but I'm not suggesting it's ready to merge--unless I guessed what would work for @hovancik and others especially well!)

### Alternative solutions

Described above :arrow_up:

### Additional information

_No response_

### Code of Conduct

lkrms:
Issue: closes #1091

### Requirements

### Description of the Change

In the `Command` class:
- Add new `--wait` option to commands `mini` and `long`.
- Add `waitToMs` method to utilise the existing interval parser (`parseDuration`) from `main.js` when processing CLI options.

`BreaksPlanner`:
- Rename `skipToMicrobreak` to `scheduleMicrobreak` and modify it slightly, adding a new `skipToMicrobreak` function that simply calls `scheduleMicrobreak(100)`
- Ditto for `skipToBreak` and `scheduleBreak`

- Fix a few minor typos and inconsistencies in logged output.

### DIFF
--- a/app/breaksPlanner.js
+++ b/app/breaksPlanner.js
@@ -168,22 +168,26 @@ class BreaksPlanner extends EventEmitter {
     this.emit('updateToolTip')
   }
 
-  skipToMicrobreak () {
+  scheduleMicrobreak (delay) {
     this.scheduler.cancel()
     const shouldBreak = this.settings.get('break')
     const shouldMicrobreak = this.settings.get('microbreak')
-    const breakInterval = this.settings.get('breakInterval') + 1
     if (shouldBreak && shouldMicrobreak) {
+      const breakInterval = this.settings.get('breakInterval') + 1
       if (this.breakNumber % breakInterval === 0) {
         this.breakNumber = 1
       }
     }
-    this.scheduler = new Scheduler(() => this.emit('startMicrobreak'), 100, 'startMicrobreak')
+    this.scheduler = new Scheduler(() => this.emit('startMicrobreak'), delay, 'startMicrobreak')
     this.scheduler.plan()
     this.emit('updateToolTip')
   }
 
-  skipToBreak () {
+  skipToMicrobreak () {
+    this.scheduleMicrobreak(100)
+  }
+
+  scheduleBreak (delay) {
     this.scheduler.cancel()
     const shouldBreak = this.settings.get('break')
     const shouldMicrobreak = this.settings.get('microbreak')
@@ -191,9 +195,13 @@ class BreaksPlanner extends EventEmitter {
       const breakInterval = this.settings.get('breakInterval') + 1
       this.breakNumber = breakInterval
     }
-    this.scheduler = new Scheduler(() => this.emit('startBreak'), 100, 'startBreak')
+    this.scheduler = new Scheduler(() => this.emit('startBreak'), delay, 'startBreak')
     this.scheduler.plan()
     this.emit('updateToolTip')
+  }
+
+  skipToBreak () {
+    this.scheduleBreak(100)
   }
 
   clear () {
--- a/app/breaksPlanner.js
+++ b/app/breaksPlanner.js
@@ -168,7 +168,7 @@ class BreaksPlanner extends EventEmitter {
     this.emit('updateToolTip')
   }
 
-  scheduleMicrobreak (delay) {
+  skipToMicrobreak (delay = 100) {
     this.scheduler.cancel()
     const shouldBreak = this.settings.get('break')
     const shouldMicrobreak = this.settings.get('microbreak')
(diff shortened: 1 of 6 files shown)


app/breaksPlanner.js
```diff
@@ -168,34 +168,42 @@ class BreaksPlanner extends EventEmitter {
...
   }
 
-  skipToMicrobreak () {
+  scheduleMicrobreak (delay) {
     this.scheduler.cancel()
     const shouldBreak = this.settings.get('break')
     const shouldMicrobreak = this.settings.get('microbreak')
-    const breakInterval = this.settings.get('breakInterval') + 1
     if (shouldBreak && shouldMicrobreak) {
+      const breakInterval = this.settings.get('breakInterval') + 1
```
hovancik, line 176: any reason why this line was moved?

app/breaksPlanner.js
```diff
@@ -168,34 +168,42 @@ class BreaksPlanner extends EventEmitter {
     this.emit('updateToolTip')
   }
 
-  skipToMicrobreak () {
+  scheduleMicrobreak (delay) {
```
hovancik: I think we should keep this function as `skipToMicrobreak (delay = 100)` instead of having 2 functions `skipToMicrobreak` and `scheduleMicrobreak`.

app/main.js
```diff
@@ -85,26 +85,39 @@ if (!gotTheLock) {
...
-        log.info('Stretchly: reseting breaks (requested by second instance)')
+        log.info('Stretchly: resetting breaks (requested by second instance)')
         resetBreaks()
         break
 
-      case 'mini':
+      case 'mini': {
         log.info('Stretchly: skip to Mini Break (requested by second instance)')
+        const ms = cmd.waitToMs()
+        if (ms === -1) {
```
hovancik: can we rename `ms` to `delay`? It would be better readable

app/main.js
```diff
@@ -942,27 +955,37 @@ function postponeBreak (shouldPlaySound = false) {
   updateTray()
 }
 
-function skipToMicrobreak () {
+function skipToMicrobreak (delay = 0) {
```
hovancik: No default value needed. (If we add it as `skipToMicrobreak (delay = 100)` in `BreaksPlanner`)

app/utils/commands.js
```diff
@@ -20,6 +20,12 @@ const allOptions = {
     description: 'Do not skip directly to this break (Long or Mini)',
     withValue: false
   },
+  wait: {
```
hovancik, line 23: Could you also please add an axample usage to `allExamples`?

hovancik (changes requested):
Thanks! I added some comments

app/breaksPlanner.js
lkrms: 1. For consistency with `skipToBreak` (I compared the functions to check for any variations in behaviour).
2. Because the call to `this.settings.get` is unnecessary unless `shouldBreak && shouldMicrobreak` is true.

app/breaksPlanner.js
lkrms: No worries. That's how I wrote it initially, but then I thought you might prefer more explicit semantics. I'll revert to the original approach.

app/main.js
lkrms: No worries. I'll make the `pause` code clearer while I'm at it, given that's where I ~~stole~~ borrowed this from :)

app/main.js
lkrms: Indeed. Actually, this was superfluous even without that. Fixed.

lkrms:
I've made the requested changes in a new commit. To avoid losing context, I didn't rebase or squash, but am happy to if needed.

hovancik (changes requested):
Thanks for the changes! I've tried locally and all seems fine :) Could you please update the CHANGELOG and squash?

hovancik (approved):

lkrms:
> Thanks for the changes! I've tried locally and all seems fine :) Could you please update the CHANGELOG and squash? 

I'm sorry I didn't get to this sooner. FOSS contribution time has been difficult to find over the last couple of weeks, but I was happy to do it. Thank you for merging. 🙏🏻

hovancik:
No worries :) I know the feeling, on the same boat here

Issue #1084, opened 2022-01-12
[Feature request]: Long brake with different color


xperiandri:
### Version

- [X] I'm using version 1.9.0

### Existing requests

- [X] I've checked Existing requests

### What operating system are you using?

Windows

### Operating System Version

Windows 11 22000

### Problem description

Sometimes you are very concentrated and skip short breaks, but I don't want to skip long breaks. So that I need to visually distinguish between one and another

### Possible solution

Add the ability to set a different color for long breaks.

### Additional information

N/A

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

hovancik:
Gotcha, that makes sense. Right now it's one value but could be separated. I would probably leave it as "advanced" option via editing config file, to keep the UI simple.

michaeldelgado1:
Hello @hovancik !

I was look though the issues that are marked as good first issues. I had a couple of questions for this one, and would like to see if I’m understanding this code enough to contribute to this issue. 

After doing a quick scan I saw this line
[https://github.com/hovancik/stretchly/blob/trunk/app/break.js#L35](https://github.com/hovancik/stretchly/blob/trunk/app/break.js#L35)

**The rest of this is assuming that this is the correct function to be interacting with for this feature :)**

Here I see we have a callback that expects to see a `backgroundColor`. Later though, I see this line:

[https://github.com/hovancik/stretchly/blob/trunk/app/break.js#L41](https://github.com/hovancik/stretchly/blob/trunk/app/break.js#L41)

This looks to be a second query of the settings file to get the `mainColor` variable. The line after seems to set a class on the body of whatever the color is minus the `#` symbol. 

Would it be possible to use `backgroundColor` here, and do the same string manipulation here as you are doing to `mainColor`. And actually, judging from how `calculateBackgroundColor` works, I don’t even think we would need to remove the `#` symbol anymore since we are doing number operations, then converting it to a string.

If my assumption is correct, we can possibly start at this line

[https://github.com/hovancik/stretchly/blob/trunk/app/main.js#L711](https://github.com/hovancik/stretchly/blob/trunk/app/main.js#L711)

We could pass some information over to `calculateBackgroundColor` in order to inform what value needs to be calculated, or even the key that we’re hoping to look up in the `settings.get()` call.

Let me know if this is close to what you had in mind, and I can get some of this written up, and put in a PR if need be.

hovancik:
Hi :) 

[...]

PR #1154, opened 2022-05-15
Making Background Colors More Flexable

michaeldelgado1:
* Fetching, and calculating background colors up front
* Reusing the cacluated background color rather than calling calculateBackgroundColor() several times
* Allowing the user to create 'breakColor' and 'microbreakColor' in their config, to choose what color they'd like for long and short breaks
* Adding backgroundColor to the classlist during 'progress' rather than querying for 'mainColor'

Issue: #1084 

### Requirements

### Other information
Here is a draft of my implementation for Issue: #1084 

There are plenty of changes that could be made here to better align with the code, but I'd like to see if this is close. I can run and write tests, as well as lint if this looks ok.

### DIFF
--- a/app/microbreak.js
+++ b/app/microbreak.js
@@ -36,8 +36,7 @@ window.onload = (e) => {
     const progressTime = document.querySelector('#progress-time')
     const postponeElement = document.querySelector('#postpone')
     const closeElement = document.querySelector('#close')
-    const mainColor = settings.get('mainColor')
-    document.body.classList.add(mainColor.substring(1))
+    document.body.classList.add(backgroundColor)
     document.body.style.backgroundColor = backgroundColor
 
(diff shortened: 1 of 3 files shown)


app/microbreak.js
```diff
@@ -36,8 +36,7 @@ window.onload = (e) => {
     const progressTime = document.querySelector('#progress-time')
     const postponeElement = document.querySelector('#postpone')
     const closeElement = document.querySelector('#close')
-    const mainColor = settings.get('mainColor')
```
hovancik, line 39: I see this is causing issue (try for example White theme), As I described here, this is not the same value: https://github.com/hovancik/stretchly/issues/1084#issuecomment-1126977211

hovancik:
Seems like no interest to work on this so closing.

PR #1195, opened 2022-09-02
Update tray only when needed

### LINKED ISSUE #1180: [Bug]:   tray icon menu UI issues on Ubuntu 20.04 Gnome
### Version

- [X] I'm using version 1.10.0

### Known issues

- [X] I've checked Known issues

### Existing issues

- [X] I've checked Existing issues

Linux

### Operating System Version

Ubuntu 20.04

### Reproduction steps

1. Install Stretchly 1.10.0
2. Start Stretchly
3. Notice that every 10 seconds, the tray icon flickers: it changes for a moment to an ellipsis (...) icon and then switches back to the regular icon.
4. Open the tray menu, open one of the submenus, for example 'skip to the next', wait for 10 seconds and notice that the submenu will close of its own.
...
**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Expected Behavior

I expect to not have an icon that flickers, or only when it actually needs to change.
I expect submenus in the tray menu to stay open until I select something or I close the menu.

### Actual Behavior

- every 10 seconds, the tray icon flickers,
- submenus in the tray menu are closed automatically after 10 seconds

### Relevant log output

```shell
None.
```

### Preferences

_No response_

### Additional information

```shell
This is caused by always calling appIcon.setImage and appIcon.setContextMenu from updateTray which is called every 10 seconds. Interestingly when I comment out one of them (doesn't matter which one), the flicker is gone.

I have a fix on my system where Stretchly remembers the previous icon path and menu template and only calls setImage when the new icon path is different and only calls setContextMenu when the new menu template is different. This fixes the issues on my machine. I can turn this into a PR if that's helpful.
```

### Code of Conduct

samegens:
Issue: closes #1180

I didn't do all the required checks yet, this is just a PR so you try it out. If you think this is the right way to fix it, I'll check everything properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - updated many translations
 - remaining times are more exact
+- only update tray when needed
 
 ## [1.11.0] - 2022-8-7
 ### Added
--- a/app/main.js
+++ b/app/main.js
@@ -56,6 +56,8 @@ let pausedForSuspendOrLock = false
 let nextIdea = null
 let appIsQuitting = false
 let updateChecker
+let currentTrayIconPath = null
+let currentTrayMenuTemplate = null
 
 require('@electron/remote/main').initialize()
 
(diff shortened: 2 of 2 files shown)


hovancik:
Hi, that looks great :) I don't see any issue with it. Feel free to complete the steps :)

hovancik (approved):

PR #1282, opened 2023-03-07
Update Linux install options

### LINKED ISSUE #703: PPA for Stretchly
### Prerequisites

- [X] I'm using latest version: v 
- [X] I've checked [Known issues](https://github.com/hovancik/stretchly#known-issues)
- [X] I've checked [Open issues](https://github.com/hovancik/stretchly/issues)

### Description

PPA for Stretchly

Operating system: 

### Steps to Reproduce

1. Go to google.com
2. Search for Stretchly PPA
3. No relevant hits
4. Same thing goes for https://github.com/hovancik/stretchly and https://hovancik.net/stretchly/

**Expected behavior:** Find a PPA for ubuntu

**Actual behavior:** No hits

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

MigVega:
closes #703 

### Requirements

### Description of the Change
Adding better description about how to install and use in Linux (tested on ubuntu 20.04).

### Verification Process
it is only two lines in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,9 @@ You can install *Stretchly* for all users silently by running this as administra
 
 ### Linux/BSD/Portable
 
-For portable versions and for Linux/BSD installers, head to [Github Releases page](https://github.com/hovancik/stretchly/releases). The most widely used distributions should be covered.
+For portable versions and for Linux/BSD installers, head to [Github Releases page](https://github.com/hovancik/stretchly/releases). The most widely used distributions should be covered. 
+Probably the best and easiest way is that you chose the instructions in  [this website](https://snapcraft.io/stretchly) (after choosing your distribution).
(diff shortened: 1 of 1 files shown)


MigVega:
Adding better description about how to install and use in Linux (tested on ubuntu 20.04).

hovancik:
Hi @MigVega , I would prefer not to recommend any single store because there are many and I am not sure which one is the best. There probably is no "best" but it probably depends on people's distros and usecases. 

What we could do is to write something like: there are many options, people can check some of them here (https://repology.org/project/stretchly/versions), there are also other stores like Flathub, Snaps...

What do you think?

hovancik (approved):

MigVega:
agree! sorry for the late reply.

Issue #1417, opened 2024-01-10
[Feature request]: migrate tests to vitest


cany748:
### Version

- [X] I'm using version 1.15.1

### Existing requests

- [X] I've checked Existing requests

### Advanced Preferences

- [X] I've checked Advanced Preferences

### What operating system are you using?

Windows

### Operating System Version

Windows 11 Pro 23H2

### Problem description

Migrate to vitest offers many benefits:
1) speed
2) ESM and TS support
3) many new features

### Possible solution

Сan migrate tests to vitest, I can create a PR. I have already migrate the tests and it works great!

### Alternative solutions

Сan use jest for tests or just leave it as is.

### Additional information

https://github.com/cany748/stretchly

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

hovancik:
What would it mean for the project setup? Right now I prefer no TS and such as I am building on CI and new dependencies usually means more issues with each new release I try to make. These dependencies tend to break random stuff and it is really painful and wasteful to spend hours trying to figure out WTF on each Stretchly release :(

cany748:
Migration to Vitest will get rid of: "chai", "chai-as-promised", "mocha", "sinon". Vitest is a flexible, well-tested, actively developing tool.

hovancik:
Well, if you say that you have done the changes already, feel free to make a PR and I will take a look

PR #1431, opened 2024-02-01
Migrate to vitest

cany748:
Issue: #1417 

### Requirements


hovancik:
Thanks! Will need to find time to check the changes, but Coverage seems to not go through

hovancik:
I am also getting `The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
`

cany748:
> I am also getting `The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details. `

There's nothing wrong with that, electron now supports ESIM and it will be easy to switch in the future

cany748:
I fixed the coverage, but there are some differences:

1. Vitest takes into account not only files in the app/utils folder, but also the entire app folder. This can be easily fixed by setting `coverage: { include: ["app/utils"] }`
2. Vitest also counts files without tests as 0% coverage. I believe this is the correct behavior, but if you disagree with me, this can also be changed via configuration.
3. We will need to create separate tests for the defaultBreakIdeas.js anddefaultMicrobreakIdeas.js files.

Otherwise, coverage is very similar

hovancik:
Thanks! I don't really care about the coverage number, I usually only look on difference when having new commits so looks good!

PR #1501, opened 2024-10-19
Recreate time in tray images

hovancik:
closes #1478

PR #1502, opened 2024-10-28
Recreate time in tray images

### LINKED ISSUE #1478: [Feature request]: More legible remaining time in tray icon
### Version

- [X] I'm using version 1.15.1

### Existing requests

### Advanced Preferences

- [X] I've checked Advanced Preferences

Windows

### Operating System Version

Windows 11 23H2

### Problem description

I have a hard time reading the number (for the remaining time to the next pause) in the tray icon.

### Possible solution

- Offer an option that does not have the Stretchly icon in the background
- Use a more legible font (e.g. make it bolder)
- Improve contrast between icon and font

### Alternative solutions

See possible solution, I added multiple options there

### Additional information

I checked the source code (too see if there is something I can do), but I did not find any script that created the many number variation icons (I guess those where created with some automation)

### Code of Conduct

Wikiwix:
Issue: closes #1478 

### Requirements

  → This is not true for the icon script: Since it is not run on the user's side I considered this okay

### Description of the Change

Starting from https://github.com/hovancik/stretchly/pull/1501 I tweaked the design and creation script of the tray icons with number in it. In detail I made the following changes (following the commit order)

#### Display more accurate remaining time in tray

- This matches the tooltip's information more closely

#### Tweak font size

The font size is now based on the baseImage's height relatively, instead of a hardcoded value.

### Verification Process

Run the dev version with over time with the new icons

### Other information

There are still two improvements I was thinking about, but the PR is already large enough as it stands…

- Make the paused icons more inline with the new numbered icons
- For the non-mac monochrome icons tweak the colours, so it matches the native colours on Windows 11 (since this is, as far as I can see the only reasonable target to optimise for (due to Mac having different icons and Linux being as diverse as it is))
- For these to changes I would switch to using the svg files as the source in the script (makes colour changes easy)


app/utils/utils.js
```diff
@@ -37,8 +37,8 @@ function formatKeyboardShortcut (keyboardShortcut) {
 }
 
 function minutesRemaining (milliseconds) {
-  const seconds = Math.ceil(milliseconds / 1000.0)
-  const minutes = Math.ceil(seconds / 60.0)
+  const seconds = Math.round(milliseconds / 1000.0)
```
hovancik: Not sure about this change. I think I prefer the original way: I took inspiration from how Apple shows time remaining in their OS installer. 

Also, there is now different logic for Icon and Tray message, so icon can show "3" and tray "Next Mini Break in about 4 minutes"

Wikiwix: > Also, there is now different logic for Icon and Tray message, so icon can show "3" and tray "Next Mini Break in about 4 minutes"

My recollection was that they were not aligned before, but I checked now, and you are correct, regarding the actual behaviour. When I look at [the code for this](https://github.com/hovancik/stretchly/blob/639088e39d23a8b8155a4c4a164bc2ec552cec03/app/utils/utils.js#L15-L23C2) though, I do not understand why this is the case 🤔 (I even read the code of the library and tested locally, which led to a rounding of the value, not the ceiling I see in the app's behaviour…)

Regarding the "taste" question:
I run it with the altered logic for ~5 days now, and I really enjoy it, because I know that a `0` in the tray means I have to prepare for the break "right now", whereas the old 1 minute did not have this urgency… (I am aware of the notifications, but I tend to not notice those, and if I glance on the icon, I am probably already not fully concentrated on anything anymore).
To put it into a single sentence: The rounding gives more accurate values for the intervals' edges, which is where it counts IMHO.

hovancik: can we then make that into separate PR? 
I don't really have hard opinions on this, but I prefer it to behave same everywhere :)

Wikiwix: Sure, I will revert it. Should I keep in the correction of the unit test?

```diff
diff --git a/test/utils.js b/test/utils.js
--- a/test/utils.js
+++ b/test/utils.js
@@ -118,10 +118,10 @@ describe('canSkip and canPostpone', () => {

   describe('minutesRemaining', () => {
     it('one minute remaining', () => {
-      minutesRemaining(1000).should.equal(1)
+      minutesRemaining(60000).should.equal(1)
     })
     it('less then one minute remaining', () => {
       minutesRemaining(1).should.equal(0)
     })
     it('ten minutes remaining', () => {
       minutesRemaining(600000).should.equal(10)
```

hovancik: I think I was trying to test that one second will show as 1 minute, which is currently correct, right? Or there is some issue with it?

Wikiwix: Yes, this is what the `one minute remaining` test checks as far as I can see.
But the name (especially taking the name of the second test `less then one minute remaining` into account) suggests (at least to me) that the test checks whether one minute input will lead to an output of one minute

hovancik: Ok, feel free to edit then 👍

Wikiwix: Pushed a commit reverting the display logic, but keeping the Unit test fix

hovancik (approved):

hovancik:
Merged, feel free to create new PR with changes to round/ceil but we need to look into how to make it consistent

app/utils/utils.js
Wikiwix: I looked into it and actually realised that there is a bug.
It is documented in #1504 and a fix is prepared in #1505

Issue #280, opened 2018-07-29
Blur effect on window for nicer integration


jancborchardt:
### Prerequisites

- [x] I'm using latest version: v0.17.0

### Description

Currently the popover is a very stark contrast to the work environment and looks quite out of place, no matter the theme.

What might be cool is a blur effect on the overlay window which nicely puts it on top of your current work. There is a package called [electron-vibrancy](https://www.npmjs.com/package/electron-vibrancy) which does just that:
![](https://i.imgur.com/42jOnRV.png)

Currently supports Windows and macOS, an issue for Linux support is open at https://github.com/arkenthera/electron-vibrancy/issues/12

hovancik:
I'll try it and add screenshot here to see what people say.

hovancik:
I'm trying from readme: 

`  electronVibrancy.SetVibrancy(mainWindow, 0);` and it exists whole app without error

`electronVibrancy.SetVibrancy(true,browserWindowInstance.getNativeWindowHandle());` gives error `window.getSize is not a function`

If anyone has more success let me know here

Sveijk:
Maybe we could add this issue to RGSoC project @hovancik ? Suggestion of @aleuca since we're now also looking into the custom changing of break background issue which is in the project.

hovancik:
:+1:

hovancik:
It is now supported in electron (for mac) but I would prefer it being the same on all platforms

https://electronjs.org/docs/api/browser-window#winsetvibrancytype-macos

hovancik:
I don't think this will get implemented so closing.

Korb:
> a blur effect on the overlay window which nicely puts it on top of your current work

Am I correct in understanding that you suggested applying a blur effect to the entire image outside the main rest notification window? That is, all currently open windows and notifications during the rest period will no longer be readable, but will be guessed. If so, that's a great suggestion.

I'll supplement it with a suggestion to add a setting to reduce the brightness of windows and the desktop outside the window notifying about the current rest.

SophiaSaiada:
I found a simple solution to implement it on MacOS (I don't know if it'll work on windows).
Is it OK to open a PR for that?

PR #1520, opened 2024-12-17
MacOS Blurred Background

SophieSaiada:
Partially solving issue #280 

Adding a simple blurred background on MacOS.
I think it should be the default behavior, but maybe it should be configurable.
It plays nice with the other configurations, such as `opacity`, `transparentMode` and `miniBreakColor`.

I also added support to opacities below 0.0392156863 (=10/255), see the comment in the Review page.

Screenshot (the buttons at the top left side of the window were like that before my change, I assume it's like that when running from source):
image

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -668,14 +668,23 @@ function startBreakNotification () {
   updateTray()
 }
 
-function getBreakWindowVibrancyOptions () {
-  if (settings.get('transparentMode') || settings.get('fullscreen') || process.platform !== 'darwin') {
+function getBlurredBackgroundWindowOptions () {
+  if (!settings.get('blurredBackground')) {
     return {}
   }
 
-  return {
-    vibrancy: 'hud',
-    visualEffectState: 'active'
+  switch (process.platform) {
+    case 'darwin':
+      return {
+        vibrancy: 'hud',
+        visualEffectState: 'active'
+      }
+    case 'win32':
+      return {
+        backgroundMaterial: 'acrylic'
+      }
+    default:
+      return {}
   }
 }
 
@@ -715,7 +724,7 @@ function startMicrobreak () {
       show: false,
       backgroundThrottling: false,
       transparent: true,
-      ...getBreakWindowVibrancyOptions(),
+      ...getBlurredBackgroundWindowOptions(),
       backgroundColor: calculateBackgroundColor(settings.get('miniBreakColor')),
       skipTaskbar: !showBreaksAsRegularWindows,
       focusable: showBreaksAsRegularWindows,
@@ -864,7 +873,7 @@ function startBreak () {
       show: false,
       backgroundThrottling: false,
       transparent: true,
-      ...getBreakWindowVibrancyOptions(),
+      ...getBlurredBackgroundWindowOptions(),
       backgroundColor: calculateBackgroundColor(settings.get('mainColor')),
       skipTaskbar: !showBreaksAsRegularWindows,
       focusable: showBreaksAsRegularWindows,
--- a/app/main.js
+++ b/app/main.js
@@ -679,10 +679,6 @@ function getBlurredBackgroundWindowOptions () {
         vibrancy: 'hud',
         visualEffectState: 'active'
       }
-    case 'win32':
-      return {
-        backgroundMaterial: 'acrylic'
-      }
     default:
       return {}
   }
--- a/app/main.js
+++ b/app/main.js
@@ -668,6 +668,17 @@ function startBreakNotification () {
   updateTray()
 }
 
+function getBreakWindowVibrancyOptions () {
+  if (settings.get('transparentMode') || settings.get('fullscreen') || process.platform !== 'darwin') {
+    return {}
+  }
+
+  return {
+    vibrancy: 'hud',
+    visualEffectState: 'active'
+  }
+}
+
 function startMicrobreak () {
   // don't start another break if break running
   if (microbreakWins) {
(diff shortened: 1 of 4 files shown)


app/main.js
```diff
@@ -1064,7 +1077,7 @@ function calculateBackgroundColor (color) {
   if (settings.get('transparentMode')) {
     opacityMultiplier = settings.get('opacity')
   }
-  return color + Math.round(opacityMultiplier * 255).toString(16)
+  return color + Math.round(opacityMultiplier * 255).toString(16).padStart(2, '0')
```
SophieSaiada, line 1085: if opacity is zero, the output will be "#ffffff0", which is an invalid color code.
The opacity part should always have exactly 2 digits.

hovancik:
Is this supported by other OSes? Seems like macOS only: https://www.electronjs.org/docs/latest/api/browser-window#winsetvibrancytype-macos

I want the app behavior to be same across platforms.

SophieSaiada:
> Is this supported by other OSes? Seems like macOS only: https://www.electronjs.org/docs/latest/api/browser-window#winsetvibrancytype-macos
> 
> I want the app behavior to be same across platforms.

Firstly, thanks for your reply and your awesome work in genral!

It's not supported by other OSes.

A cross-platform solution will be much more complicated.
It will probably include an external library and won't be a native solution, resulting on excessive resources usage for that blurring effect.

Stretchly already has a non cross-platform feature - do-not-disturb doesn't work on MacOS.

I recently found a similar paid app, called LookAway. One of their two strengths, compared to Stretchly, is that they have this beautiful blur effect, and that they have a good meetings detection.
The second feature will be far more complicated to implement/fix, so I try to at least level Stretchly up with them in the visual aspect.

Anyway, I don't have a PC or a linux-based computer, so I won't be able to code/test other OSes.

hovancik:
Well DND was supported until Apple changed things and I am not able to detect it anymore. 

Anyway, I think if we have this as [advanced option](https://github.com/hovancik/stretchly?tab=readme-ov-file#advanced-preferences) with default to `false`, I can merge the PR.  Seems like for windows they have similar with `acrylic` [win.setBackgroundMaterial(material)](https://www.electronjs.org/docs/latest/api/browser-window#winsetbackgroundmaterialmaterial-windows)

SophieSaiada:
Ok, fair point :)

An advanced option sounds amazing, I'll work on it on the weekend.

Sure, I can try using the counterpart Windows' option, but I'll need someone else to test it.

hovancik:
Great, if you can try adding Windows as well, I can test it. I will see if anything can be done on Linux as well.

SophieSaiada:
I made it an advanced option, and also added Windows support that should be tested. :)

hovancik:
Checking on Windows and it doesn't work :( Seems like there is bug in electron: https://github.com/electron/electron/issues/43345 . In my experience, I am very skeptical it will ever be fixed :(

Maybe remove windows code and only keep this mac-only, of it works there? Not sure how to proceed, if it makes sense to add windows code that does not work.

SophieSaiada:
Yes, I had a feeling it won't be that easy 😵‍💫
I removed the Windows code, I think it makes more sense than leaving a code that's not working.

hovancik (approved):
Thanks!

vardecab:
@SophiaSaiada, are you sure this didn't brake the `opacity` feature? Everything was fine on my end up until I updated to the latest version of the app. I've noticed the opacity setting doesn't work so I started digging and I found this issue/PR. 

New blur option doesn't work either. Do both `blurredBackground` **and** `transparentMode` need to be set to `true` for blur to work?

No matter what true/false combo I use, the brake window is always fully opaque 👀

App version 1.17.1.

image

UQRMvvi6

hovancik:
I am seeing the same, strange. Does it work for you @SophiaSaiada ?

SophieSaiada:
Weird, it works fine for me (macbook pro M3, macOS 15.2).

image

image

  
Maybe [this line](https://github.com/hovancik/stretchly/pull/1520/files#diff-5f5ac547707d85825f5ce55d9deb200422b29d711b82263eb4bd3a1cab6a661aR1085) is the problem? I can't see the bug, but it's the only line that can affect the opacity (it adds support in lower opacity values).



I can open a PR that reverts this line, and then you could build and test this branch and see if it solves the issue. What do you think?

hovancik:
@vardecab what hardware do you use? I think the issue is related to newer Electron and older hardware. 

On my superold macbook even removing the `pad` does not help. @SophiaSaiada

vardecab:
> @vardecab what hardware do you use? I think the issue is related to newer Electron and older hardware.
> 
> On my superold macbook even removing the `pad` does not help. @SophiaSaiada

MacBook Pro M3, macOS 15.2

The behavior is highly inconsistent — on one break it works fine, but then for two consecutive breaks it fails and reverts to fully opaque. Then it works again, only to stop working once more :/ 

Additionally, when connected to an external display, one screen (external) often shows as opaque while the other (built-in) appears blurred:
37811d98ec69d10afb357e11d8e91f5e01c4be532235df485b57a3d5b9ff07e0

Those are my settings in `config.json` FWIW (I want the blurred version that shows what's "under" the break screen so I think both `transparentMode` and `blurredBackground` need to be `true` as well as `opacity` set to something low such as `0.1` — maybe I'm wrong here?):
 
```"mainColor": "#707070",
	"miniBreakColor": "#707070",
	"transparentMode": true,
	"blurredBackground": true,
	"opacity": 0.1,
```

PR #1599, opened 2025-05-01
Move to ESM

hovancik:


devtobi:
@hovancik Any updates on when this refactoring will be finished? Since this looks like a major rewrite do you plan to release a new major version?

hovancik:
@devtobi it changes a lot of things, but it's mostly migration from require to import, related changes and some things I catched in between related to new electron version and such. 

So user will not see any changes. (hopefully)

I am pretty happy with current state, so I think I will merge it in coming days and see. I am trying to run `trunk` where I can so hopefully there won't be any regression I am not catching

Issue #1596, opened 2025-04-17
[Feature request]: Mini-Breaks - Separate physical and psychological ideas


imagoiq:
### Version

- [x] I'm using version 1.17.2

### Existing requests

- [x] I've checked Existing requests

### Advanced Preferences

- [x] I've checked Advanced Preferences

### What operating system are you using?

macOS

### Operating System Version

macOS Sequoia 15.3.2

### Problem description

I would love to be able to see mini breaks ideas only if the ideas are about doing a physical exercise and avoid displaying more ideas with psychology involved (e.g. "Take a moment to smile at being alive." https://github.com/hovancik/stretchly/blob/trunk/app/locales/en.json#L255)

### Possible solution

I guess, add a `category` field in the locales, add JSON parameter in the config and a condition somewhere

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

hovancik:
Hi and thanks for the feedback. I am thinking for version 2 (TBH probably far future), I would like to have something like this: pick different kind of breaks that user can choose to show/not show.

ptandler:
Hi, I also was thinking about something in this direction for quite some time now: Add a category to each idea and then allow users to select probabilities for each category, e.g. 70 % physical, 20 % psychology and 10 % custom ... or alike, you get the idea.

Maybe also improve the algorithm for the selection of the idea to show at a break to ensure to cover all categories (with the configured probability) before repeating and ensure that the ideas in each category will be shown all before repeating.

And in addition with preference profiles (https://github.com/hovancik/stretchly/issues/1595) you could easily configure to use a different set of suggestions for the office and home-office environments 🙂

I'm not sure if I really can invest some spare time, but I'd really love to start a PR for this.

ptandler:
what about these categories:

- Physical Exercise: Stretching, movement, posture, desk exercises
- Eye Care: Blinking, focus change, exposure to natural light
- Mindfulness & Relaxation: Breathing, meditation, visualization, gratitude
- Social & Collaboration: Breaks with co-workers, communication, active meetings
- Ergonomics & Workspace: Desk setup, decluttering, changing sitting posture
- Healthy Habits: Hydration, eating fruit, taking stairs, walking
- Creativity & Mental Stimulation: Art, learning, imaginative tasks

ptandler:
[...]

PR #1669, opened 2025-10-09
add categories for break ideas

ptandler:
Issue: #1596

### Requirements

### Description of the Change

As first step, I added 7 categories as proposed in https://github.com/hovancik/stretchly/issues/1596#issuecomment-3386199302

I'd suggest to use the idea ID's prefix to identify the category, as this makes handling the ideas easier and we don't need to maintain an additional prop for the category.

Next step would be to improve IdeasLoader & Shuffed

### DIFF
--- a/app/locales/en.json
+++ b/app/locales/en.json
@@ -212,304 +212,326 @@
     "resuming": "Resuming"
   },
   "miniBreakIdeas": {
-    "aaa": {
-      "text": "Go grab a glass of water."
-    },
-    "aab": {
-      "text": "Slowly look all the way left, then right."
-    },
-    "aac": {
-      "text": "Slowly look all the way up, then down."
-    },
-    "aad": {
-      "text": "Close your eyes and take few deep breaths."
-    },
-    "aae": {
-      "text": "Close your eyes and relax."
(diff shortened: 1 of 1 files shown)


steveszat:
I like the direction here — adding structure to break ideas makes a lot of sense.

Before getting into implementation details, I stepped back and thought about how I’d approach this if I were designing it from scratch.

To me, break ideas feel more like content than general application settings. So, I’d probably think in terms of:

* a message library (the ideas themselves)
* categories (eye relief, stretching, focus, etc.)
* user preferences (which categories are enabled)
* scheduling and presentation as separate concerns

In that model, messages would be structured objects with category metadata, and managed somewhat independently from the rest of the app configuration.

Given where things are today, it seems like we can still move toward something like this incrementally.

One idea would be to introduce a small abstraction for loading break ideas instead of reading them directly from settings. That would make it easier to:

* normalize message structure
* add categories cleanly
* evolve the storage later if needed

I’m fully on board with adding categories — this just seems like a way to make the change easier to extend and maintain over time.

Curious how you’re thinking about it, or if you have a different approach in mind.

PR #1670, opened 2025-10-11
Add flatpak portal manager, autostart on flatpaks

### LINKED ISSUE #1652: [Feature request]: Autostart for Flatpaks
### Version

- [x] I'm using version 1.18.0

### Existing requests

- [x] I've checked Existing requests

### Advanced Preferences

- [x] I've checked Advanced Preferences

Linux

### Operating System Version

Fedora 42, Aurora

### Problem description

Autostart doesn't work when its in a flatpak, especially on an immutable OS like Aurora.

This has been noted in #1517 

### Possible solution

I'd like to use Flatpak Portals to request the program starts at login

This is documented in [Background Portals](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Background.html)

I have an implementation with Claude. But to test it, I need to package it as a Flatpak, instead of just `npm run dev`

---
I have significant issues using the `flatpak-builder` command, trying to use the manifest at [net.hovancik.Stretchly](https://github.com/flathub/net.hovancik.Stretchly)

1. Issue with generated-sources. So I copy `package-lock.json` from the main stretchly repo & rebuild generated sources with the `flatpak-node-generator`
2. Update manifest from `node20` to `node22` which Stretchly now uses
3. Esbuild binary error (no idea how to resolve, been trying to add things to the manifest but it doesn't help)
`Error: ENOENT: no such file or directory, stat '/run/build/stretchly/flatpak-node/cache/esbuild/bin/esbuild-current'`

Sorry if any of this was incorrect. I'd appreciate any help. Thanks for this program 🙂 

### Alternative solutions

I'm pretty sure the only way out of a Flatpak is with the portals

### Additional information

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Zinkelburger:
Issue: closes #1652 #1517 

### Requirements

### Description of the Change

1. Add Flatpak detection to AutoStart manager
2. Changed AutoStart manager to async because d-bus methods are async
3. Follow docs at [doc-org.freedesktop.portal.Background](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Background.html) for `options` field in `setAutostart`
4. Get background status by checking for `~/.config/autostart/net.hovancik.Stretchly.desktop`. Requires a PR to the flatpak repo. https://github.com/flathub/net.hovancik.Stretchly/pull/23

### Verification Process

I was on Fedora 42.
I build the flatpak and ran it as described in [net.hovancik.Stretchly](https://github.com/flathub/net.hovancik.Stretchly)

I checked the checkbox to "run at startup", and saw that a file was made in autostart, as expected.
```
cat ~/.config/autostart/net.hovancik.Stretchly.desktop    
[Desktop Entry]  
Type=Application  
Name=net.hovancik.Stretchly  
X-XDP-Autostart=net.hovancik.Stretchly  
Exec=flatpak run net.hovancik.Stretchly  
X-Flatpak=net.hovancik.Stretchly
```

When I unchecked the checkbox, the file no longer existed. Perfect.
When I checked the checkbox and rebooted my computer, stretchly opened on startup.

I fixed the error I spoke of previously

Log:
```
12:07:48.985 › Stretchly: setting autostart to true on linux (Flatpak)
12:07:48.991 › Stretchly: RequestBackground called for autostart=true, request path: /org/freedesktop/portal/desktop/request/1_266/stretchly_autostart_1760198868989_s3uwx
12:07:48.991 › Stretchly: Portal Response signal received - response: 0, results: {
  background: { signature: 'b', value: true },
  autostart: { signature: 'b', value: true }
}
12:07:48.993 › Stretchly: Autostart successfully enabled via XDG Portal.
12:07:49.951 › Stretchly: setting autostart to false on linux (Flatpak)
12:07:49.958 › Stretchly: Autostart disabled via XDG Portal (no response expected).
12:07:50.582 › Stretchly: setting autostart to true on linux (Flatpak)
12:07:50.590 › Stretchly: RequestBackground called for autostart=true, request path: /org/freedesktop/portal/desktop/request/1_266/stretchly_autostart_1760198870587_qae8dg
12:07:50.590 › Stretchly: Portal Response signal received - response: 0, results: {
  background: { signature: 'b', value: true },
  autostart: { signature: 'b', value: true }
}
12:07:50.591 › Stretchly: Autostart successfully enabled via XDG Portal.
```

### DIFF
--- a/app/utils/autostartManager.js
+++ b/app/utils/autostartManager.js
@@ -14,7 +14,7 @@ class AutostartManager {
     this.windowsStore = windowsStore
     this.app = app
 
-    this.isFlatpak = this.platform === 'linux' && insideFlatpak()
+    this.isFlatpak = insideFlatpak()
 
     if (this.isFlatpak) {
       this.flatpakPortalManager = new FlatpakPortalManager(settings)
--- a/app/utils/autostartManager.js
+++ b/app/utils/autostartManager.js
@@ -1,30 +1,54 @@
 import log from 'electron-log/main.js'
 import AutoLaunch from 'auto-launch'
+import FlatpakPortalManager from './flatpakPortalManager.js'
+import { insideFlatpak } from './utils.js'
 
 class AutostartManager {
   constructor ({
     platform,
     windowsStore,
-    app
+    app,
+    settings
   }) {
     this.platform = platform
     this.windowsStore = windowsStore
     this.app = app
+    this.settings = settings
+    this.flatpakPortalManager = new FlatpakPortalManager()
   }
 
-  setAutostartEnabled (value) {
-    if (this.platform === 'linux') {
+  async setAutostartEnabled (value) {
+    log.info(`Stretchly: setting autostart to ${value} on ${this.platform}${this.platform === 'win32' && this.windowsStore ? ' (Windows Store)' : ''}${insideFlatpak() && this.platform === 'linux' ? ' (Flatpak)' : ''}`)
+    if (this.platform === 'linux' && insideFlatpak()) {
+      try {
+        // Initialize portal manager first
+        await this.flatpakPortalManager.initialize()
+
+        const result = await this.flatpakPortalManager.setAutostart(value)
+
+        // Only save to settings if portal call succeeded
+        if (result) {
+          this.settings.set('flatpakAutostart', value)
+          log.info(`Stretchly: Saved flatpakAutostart=${value} to settings after successful portal call`)
+        } else {
+          log.warn('Stretchly: Portal call returned false, not saving flatpakAutostart setting')
+        }
+      } catch (error) {
+        log.error('Stretchly: Failed to set autostart via XDG Portal. No fallback available for Flatpak.', error)
+      }
+    } else if (this.platform === 'linux') {
       value ? this._linuxAutoLaunch.enable() : this._linuxAutoLaunch.disable()
     } else if (this.platform === 'win32' && this.windowsStore) {
       value ? this._windowsStoreAutoLaunch.enable() : this._windowsStoreAutoLaunch.disable()
     } else {
       this.app.setLoginItemSettings({ openAtLogin: value })
     }
-    log.info(`Stretchly: setting autostart to ${value} on ${this.platform}${this.platform === 'win32' && this.windowsStore ? ' (Windows Store)' : ''}`)
   }
 
   async autoLaunchStatus () {
-    if (this.platform === 'linux') {
+    if (this.platform === 'linux' && insideFlatpak()) {
+      return this.settings.get('flatpakAutostart', false)
+    } else if (this.platform === 'linux') {
       return await this._linuxAutoLaunch.isEnabled()
     } else if (this.platform === 'win32' && this.windowsStore) {
       return await this._windowsStoreAutoLaunch.isEnabled()
@@ -47,6 +71,12 @@ class AutostartManager {
       isHidden: true
     })
     return stretchlyAutoLaunch
+  }
+
+  disconnect () {
+    if (this.flatpakPortalManager) {
+      this.flatpakPortalManager.disconnect()
+    }
   }
 }
 
--- a/app/utils/autostartManager.js
+++ b/app/utils/autostartManager.js
@@ -20,6 +20,12 @@ class AutostartManager {
       this.flatpakPortalManager = new FlatpakPortalManager(settings)
     } else if (this.platform === 'linux') {
       this.nativeAutoLauncher = new AutoLaunch({ name: 'stretchly' })
+    } else if (this.platform === 'win32' && this.windowsStore) {
+      this.windowsStoreAutoLauncher = new AutoLaunch({
+        name: 'Stretchly',
+        path: '33881JanHovancik.stretchly_24fg4m0zq65je!Stretchly',
+        isHidden: true
+      })
     }
   }
 
@@ -31,7 +37,7 @@ class AutostartManager {
     } else if (this.platform === 'linux') {
       await (value ? this.nativeAutoLauncher.enable() : this.nativeAutoLauncher.disable())
     } else if (this.platform === 'win32' && this.windowsStore) {
-      await (value ? this._windowsStoreAutoLaunch.enable() : this._windowsStoreAutoLaunch.disable())
+      await (value ? this.windowsStoreAutoLauncher.enable() : this.windowsStoreAutoLauncher.disable())
     } else {
       this.app.setLoginItemSettings({ openAtLogin: value })
     }
@@ -43,19 +49,10 @@ class AutostartManager {
     } else if (this.platform === 'linux') {
       return await this.nativeAutoLauncher.isEnabled()
     } else if (this.platform === 'win32' && this.windowsStore) {
-      return await this._windowsStoreAutoLaunch.isEnabled()
+      return await this.windowsStoreAutoLauncher.isEnabled()
     } else {
-      return await this.app.getLoginItemSettings().openAtLogin
+      return this.app.getLoginItemSettings().openAtLogin
     }
-  }
-
-  get _windowsStoreAutoLaunch () {
-    const stretchlyAutoLaunch = new AutoLaunch({
-      name: 'Stretchly',
-      path: '33881JanHovancik.stretchly_24fg4m0zq65je!Stretchly',
-      isHidden: true
-    })
-    return stretchlyAutoLaunch
   }
 
   disconnect () {
--- a/app/utils/autostartManager.js
+++ b/app/utils/autostartManager.js
@@ -13,50 +13,23 @@ class AutostartManager {
     this.platform = platform
     this.windowsStore = windowsStore
     this.app = app
-    this.settings = settings
-    this.flatpakPortalManager = new FlatpakPortalManager()
 
-    // Decide the Linux autostart strategy once during construction
-    if (this.platform === 'linux') {
-      if (insideFlatpak()) {
-        this._linuxAutoLaunch = this._createFlatpakAutostarter()
-      } else {
-        this._linuxAutoLaunch = new AutoLaunch({ name: 'stretchly' })
-      }
-    }
-  }
+    this.isFlatpak = this.platform === 'linux' && insideFlatpak()
 
-  _createFlatpakAutostarter () {
-    return {
-      enable: async () => {
-        try {
-          await this.flatpakPortalManager.initialize()
-          await this.flatpakPortalManager.setAutostart(true)
-          this.settings.set('flatpakAutostart', true)
-        } catch (error) {
-          log.error('Stretchly: Failed to set autostart (enable) via XDG Portal', error)
-        }
-      },
-      disable: async () => {
-        try {
-          await this.flatpakPortalManager.initialize()
-          await this.flatpakPortalManager.setAutostart(false)
-          this.settings.set('flatpakAutostart', false)
-        } catch (error) {
-          log.error('Stretchly: Failed to set autostart (disable) via XDG Portal', error)
-        }
-      },
-      isEnabled: async () => {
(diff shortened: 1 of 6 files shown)


hovancik:
Is this WIP or should be finished? With regard to the PR in flathub, if any changes are expected

Zinkelburger:
That's a good point, the flathub PR is definitely wrong, that permission is not allowed.

I'll rethink how to do this, I had a lot of errors trying to get the current status, and reading "autostart" was a good workaround.

hovancik:
Ok :)

Zinkelburger:
There were issues with the portal not initializing sometimes, so I moved it to main, hope that's ok.

Otherwise its pretty much the last PR + save autostart value to settings. Saving to settings is kinda buggy if the user manually deletes the file, but they're asking for trouble in that case!

app/utils/autostartManager.js
```diff
@@ -1,30 +1,54 @@
...
+    if (this.platform === 'linux' && insideFlatpak()) {
+      try {
+        // Initialize portal manager first
+        await this.flatpakPortalManager.initialize()
+
+        const result = await this.flatpakPortalManager.setAutostart(value)
+
+        // Only save to settings if portal call succeeded
+        if (result) {
+          this.settings.set('flatpakAutostart', value)
```
hovancik: Why do we need to save in preferences?

app/utils/flatpakPortalManager.js
```diff
@@ -0,0 +1,158 @@
...
+
+            if (response === 0) { // Success
+              const autostartGranted = results && results.autostart && results.autostart.value === enabled
+              if (autostartGranted) {
+                log.info('Stretchly: Autostart successfully enabled via XDG Portal.')
+              } else {
+                log.warn('Stretchly: Autostart status did not match request. Results:', results)
+              }
+              resolve(autostartGranted)
+            } else if (response === 1) { // User cancelled
```
hovancik: How can user cancel?

app/utils/autostartManager.js
```diff
@@ -1,30 +1,54 @@
...
     this.app = app
+    this.settings = settings
+    this.flatpakPortalManager = new FlatpakPortalManager()
   }
 
-  setAutostartEnabled (value) {
-    if (this.platform === 'linux') {
+  async setAutostartEnabled (value) {
+    log.info(`Stretchly: setting autostart to ${value} on ${this.platform}${this.platform === 'win32' && this.windowsStore ? ' (Windows Store)' : ''}${insideFlatpak() && this.platform === 'linux' ? ' (Flatpak)' : ''}`)
+    if (this.platform === 'linux' && insideFlatpak()) {
```
hovancik: I think we should differentiate between flatpak and not flatpak in `_linuxAutoLaunch`

app/utils/flatpakPortalManager.js
Zinkelburger: I think there can be a dialog on some desktop environments, but not on any of the Fedora machines I've used it on.
https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Request.html

app/utils/autostartManager.js
Zinkelburger: There doesn't appear to be a way to query the background portal, only to set and unset the value. So I save the value instead.

When I try to query the status with no value in autostart, it appears to be unsetting it.

E.g. with no autostart key set:

```
      const options = {
        handle_token: new Variant('s', handleToken),
        reason: new Variant('s', 'Stretchly is checking its background and autostart permissions.')
      }
```

returns
```
16:04:51.330 › Stretchly: RequestBackground (query) called, request path: /org/freedesktop/portal/desktop/request/1_294/stretchly_query_1762463091326_959g6
16:04:51.333 › Stretchly: Portal Query Response received - response: 0 {
  background: { signature: 'b', value: true },
  autostart: { signature: 'b', value: false }
}
```
Even when autostart was set.

app/utils/autostartManager.js
Zinkelburger: That's a very good point

app/utils/autostartManager.js
```diff
@@ -1,26 +1,67 @@
...
+    app,
+    settings
   }) {
     this.platform = platform
     this.windowsStore = windowsStore
     this.app = app
+    this.settings = settings
+    this.flatpakPortalManager = new FlatpakPortalManager()
+
+    // Decide the Linux autostart strategy once during construction
```
hovancik: I don't understand this change. Why is this here?

app/utils/autostartManager.js
```diff
@@ -1,26 +1,67 @@
...
+          await this.flatpakPortalManager.initialize()
+          await this.flatpakPortalManager.setAutostart(false)
+          this.settings.set('flatpakAutostart', false)
+        } catch (error) {
+          log.error('Stretchly: Failed to set autostart (disable) via XDG Portal', error)
+        }
+      },
+      isEnabled: async () => {
+        // XDG portals don't provide a reliable query method, so we read from our cache
+        return Promise.resolve(this.settings.get('flatpakAutostart', false))
```
hovancik: Why we have promise.resolve here? and false in get?

app/utils/autostartManager.js
```diff
@@ -1,26 +1,67 @@
...
+        this._linuxAutoLaunch = new AutoLaunch({ name: 'stretchly' })
+      }
+    }
   }
 
-  setAutostartEnabled (value) {
+  _createFlatpakAutostarter () {
+    return {
+      enable: async () => {
+        try {
```
hovancik: I personally don't like try catch, as I don't really then know what's happening. Is it needed here (and elsewhere)?  If it fails often, I would prefer we prevent it from failing.

app/utils/autostartManager.js
hovancik: Got it, you then try to follow how **AutoLaunch** behaves. Can we actually make it this way: 
- flatpak manager is a single file / class that does everything and  in here we just call it?

I think the mental overhead will be smaller then to realize that we copy some other behaviour (and we could change/remove/add another linux way in the future). So some `if`s are better for me than mental overhad.  

Hope this makes sense.

app/utils/autostartManager.js
Zinkelburger: That's a good point, but I think these are necessary. For example, if the `xdg-desktop-portal` isn't installed, it could throw an error. Anything with the D-bus could potentially fail, its an OS problem, not in our control.

I did remove the re-throwing of these errors, now they're just logged. And refactored it only FlatpakPortalManager has try-catch blocks.

app/utils/autostartManager.js
Zinkelburger: That's true, I've removed the Promise.resolve.

False in get because it might not exist in the config.
I updated it to initialize to false in defaultSettings.js instead

app/utils/autostartManager.js
Zinkelburger: Yes that makes a lot of sense,  I should handle all the flatpak portal manager stuff in one file instead of two _autoLaunch instances things

Zinkelburger:
Please note I have an issue running the flatpak with no flags on stretchly 1.19.0. I think there were some changes to the dependencies.
```
anbernal@anbernal-thinkpadt14sgen2i:~$ flatpak run net.hovancik.Stretchly
WAYLAND_DISPLAY: wayland-0
XDG_RUNTIME_DIR: /run/user/4215325
app  bus  doc  flatpak-info  p11-kit  pulse
[4:1117/002546.213869:ERROR:dbus/bus.cc:408] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[4:1117/002546.958812:ERROR:ui/ozone/platform/wayland/host/wayland_connection.cc:202] Failed to connect to Wayland display: No such file or directory (2)
[4:1117/002546.958846:ERROR:ui/ozone/platform/wayland/ozone_platform_wayland.cc:282] Failed to initialize Wayland platform
[4:1117/002546.958870:ERROR:ui/aura/env.cc:257] The platform failed to initialize.  Exiting.
/app/bin/run.sh: line 14:     4 Segmentation fault      (core dumped) zypak-wrapper /app/main/stretchly $FLAGS "$@"
```
To get it to work I have to add the --ozone-platform=x11
```
anbernal@anbernal-thinkpadt14sgen2i:~$ flatpak run net.hovancik.Stretchly --ozone-platform=x11
WAYLAND_DISPLAY: wayland-0
XDG_RUNTIME_DIR: /run/user/4215325
app  bus  doc  flatpak-info  p11-kit  pulse
[4:1117/003950.727471:ERROR:dbus/bus.cc:408] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[4:1117/003952.056727:ERROR:dbus/bus.cc:408] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[4:1117/003952.056822:ERROR:dbus/bus.cc:408] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
00:39:52.089 › Stretchly: initializing...
00:39:52.129 › Stretchly: migrating preferences from Stretchly v1.18.1 to v1.18.2
00:39:52.144 › Stretchly: migrating preferences from Stretchly v1.18.2 to v1.19.0
00:39:52.145 › Stretchly: not migrating audio to longBreakAudio
```

I'm not an expert and its late at night. Claude suggests 
"Electron 39 changed the default Ozone platform behavior. You used to need to explicitly set `--enable-features=UseOzonePlatform --ozone-platform=wayland`. The default is now `auto`, and it tries to use Wayland when
  `XDG_SESSION_TYPE=wayland`. But the Flatpak sandbox doesn't have `--socket=wayland` permission, so the socket isn't accessible"

hovancik:
Yeah, that is a change in Electron :( Are you on Ubuntu? looks like issue in #1693

hovancik:
Made some changes, but one question: this is supposed to be triggered from Preferences. But I do not see where we actually load the value into Preferences. Or am I missing something?

app/utils/autostartManager.js
```diff
@@ -1,45 +1,54 @@
...
+    } else if (this.platform === 'linux') {
+      this.nativeAutoLauncher = new AutoLaunch({ name: 'stretchly' })
+    }
   }
 
-  setAutostartEnabled (value) {
-    if (this.platform === 'linux') {
-      value ? this._linuxAutoLaunch.enable() : this._linuxAutoLaunch.disable()
+  async setAutostartEnabled (value) {
+    log.info(`Stretchly: setting autostart to ${value} on ${this.platform}${this.platform === 'win32' && this.windowsStore ? ' (Windows Store)' : ''}${this.isFlatpak ? ' (Flatpak)' : ''}`)
```
hovancik: is there some reason why this was moved to the begginning of the function?

Zinkelburger:
> Made some changes, but one question: this is supposed to be triggered from Preferences. But I do not see where we actually load the value into Preferences. Or am I missing something?

The value is loaded via the flatpakAutostart setting (since we must cache it locally; the portal doesn't support querying status).

It gets injected into the Preferences window in app/main.js via settingsToSend():
```javascript
async function settingsToSend () { 
  return Object.assign({}, settings.store, { openAtLogin: await autostartManager.autoLaunchStatus() })
}
```
This maps our cached flatpakAutostart value (returned by autoLaunchStatus) to the openAtLogin property that the Preferences UI expects.

app/utils/autostartManager.js
Zinkelburger: "Setting autostart" implies it's going to happen in the future. If it was logged as "set autostart", then it would make sense to have at the end of the function.

Also if one of the operations fails, we can still see that the attempt was made. In my debugging it was more useful to have it at the beginning than the end.

hovancik:
looks like you deleted my changes :(

Issue #1604, opened 2025-05-09
[Bug]:   Consuming too much memory.


iamsoum:
### Version

- [x] I'm using version 1.17.2

### Known issues

- [x] I've checked Known issues

### Existing issues

- [x] I've checked Existing issues

### Advanced Preferences

- [x] I've checked Advanced Preferences

### What operating system are you using?

macOS

### Operating System Version

Mac OS Sequoia 15.4.1

### Reproduction steps

1. Just use Stretchly normally
2. You get the too much RAM usage alert at random times
3. Check your Activity monitor for double checking
...
**Reproduces how often:** 50%

### Expected Behavior

Expected behavior would be 200 MB of memory consumption or something.

### Actual Behavior

<img width="1702" alt="Image" src="https://github.com/user-attachments/assets/46803a6c-0092-48e7-95f6-2e08a1eb231b" />

<img width="393" alt="Image" src="https://github.com/user-attachments/assets/69e4ee31-9c7a-4467-a627-cc5864e15781" />

### Relevant log output

```shell
I don't know how to fetch the log messages from Sretly.
```

### Preferences

```shell
I have default preferences.
```

### Additional information

I'm on M2 MacBook Air.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

vysakh-poulose:
You can fetch the log messages by using the `Ctrl/Cmd + D` shortcut while viewing the About section of Preferences. Check the second paragraph [here](https://github.com/hovancik/stretchly#advanced-preferences)

<img width="680" alt="Image" src="https://github.com/user-attachments/assets/50791331-66c4-4e29-8c35-65fd147aaa3d" />

924067327:
i have the same question too

924067327:
<img width="1020" height="646" alt="Image" src="https://github.com/user-attachments/assets/e678af8d-b67d-40f9-8ad1-e818a6e41543" />

jhult:
<img width="778" height="537" alt="Image" src="https://github.com/user-attachments/assets/1d494b52-131c-4254-9d88-3c614698a79f" />

jhult:
Possibly related: https://github.com/hovancik/stretchly/issues/1278

hovancik:
can you folks check if the number of processes is rising or stays the same? Electron opens multiple processes, but they should stay the same. Except when in Break, new process gets created but should be closed after. 

Unfortunatelly, I do not have mac.

jhult:
[...]

PR #1700, opened 2025-11-27
fix: break windows not closing correctly on all platforms

jhult:
Issue: #1604

### Description of the Change

- Removed 1 global variable, 5 lines from `breakComplete()`, 2 checks from event handlers
- Strict mode still prevents users from closing windows

### Benefits

- More robust: `destroy()` works regardless of platform or window properties
- Clearer intent: Programmatic `destroy()` vs. user-prevented `close()`
- Works everywhere: Consistent behavior across macOS/Linux/Windows

#### User tries to close break window
  - `close` event fires → handler checks strict mode → `preventDefault()` blocks it

#### Programmatically closing (finish/postpone/skip)
  - `window.destroy()` called → closed event fires → cleanup happens
  - Bypasses the `close` event entirely (intentional!)

### Verification Process

1. `npm run dist -- --mac --arm64`
2. Install new build in Applications and run for a few hours with regular breaks
3. Check Activity Monitor - confirm no accumulation of renderer processes
4. Check memory usage - should stay < 200 MiB, not grow over time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Fixed
 - snap package not starting on Wayland
+- break windows not closing correctly on all platforms
 
 ## [1.19.0] - 2025-11-13
 ### Added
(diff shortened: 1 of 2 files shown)


hovancik (approved):
LGTM 🥇

PR #1703, opened 2025-11-30
Adds new icon styles preference for tray

### LINKED ISSUE #1558: [Feature request]: Progress bar indicator in tray
### Version

- [x] I'm using version 1.17.2

### Existing requests

- [x] I've checked Existing requests

### Advanced Preferences

- [x] I've checked Advanced Preferences

Other (specify below)

### Operating System Version

Windows 11 Version 22H2 (OS Build 19045.4651)

### Problem description

I would like to be able to tell at a glance what is the remaining time till next short and long break. The existing setting to show a number in tray icon does not satisfy me, as it only shows one of those, plus I do not find it particularly readable at small size.

### Possible solution

I think the best solution would be to show two circular progress bars, one for short and one for long break. E.g. if there is 4/5 minutes left to short break and 5/50 minutes to long break, they should show 20% and 90% respectively (time elapsed = 1 - time remaining).

It could also be used to show remaining pause time, though if I recall correctly the app currently does not remember how long the pause was, only how long much longer will it take. 

### Alternative solutions

Some consideration should be taken for particular designs to make both readable at tray icon size:
- Should they be one within another like Apple Watch fitness trackers?
- Or maybe partially overlap to make them both bigger?
- Maybe something more like a clock, with smaller and bigger hands?
- Or make the long break a circular progress bar, with remaining circle in the middle filling from bottom to top as short break progresses?
- Or maybe use horizontal/vertical bars instead?

### Additional information

If the used framework does not play well with generating progress bars on the fly, pre-generating icons with all the possible combinations at some step size can be used instead, either at compile time or whenever changing settings.

E.g. With 5 minute short breaks and 50 minute long breaks, at 1-minute resolution, we have 50 possible icons. The short break break length does not really play into the number here, but it may affect chosen resolution so that it is always a whole number of steps (so for 6 minute short breaks, we could have chosen 2-minute resolution instead). 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

hovancik:
closes #1558


Wikiwix:
@hovancik I saw this change by chance today: As far as I see it, this PR reintroduced Issue #1478. Is this intentional?

hovancik:
what's the OS/theme you use? for the ones I use new icons are improvements

Wikiwix:
I was only looking at this PR's diff, not the icon in the running program.
Not sure what I saw though. Cannot find what I thought I saw earlier and can confirm that 1.20.0 still has legible text (while looking different).
Sorry for the inconvenience 😥

---

As it is connected to this PR: The inverted monochrome icon colour does not match very well on windows 11 (most visible in the new Progress style, but also with the number as text).
(The Windows icons are `#FFFFFF`)

Windows 11 Tray with monochrome Stretchly icon

hovancik:
Maybe let's create an issue for this? Seems similar for my Debian KDE, but no idea how about other OSes - whether we should have white everywhere

PR #1729, opened 2026-01-21
Fix focus mode detection on macOS Tahoe

zhekai-jiang:
Issue: #1728 

### Requirements

- [ ]  all new code has tests to ensure against regressions. – I suppose it's tricky to write a unit test for this feature as it is OS-specific. I see no unit test that actually checks this functionality in [test/dndManager.js](https://github.com/hovancik/stretchly/blob/trunk/test/dndManager.js) yet.

### Description of the Change

On macOS, we now check the version of the system and apply different commands to check for focus modes.

### Verification Process

Actions:
1. Uncheck "Show breaks even in Do Not Disturb mode" in preferences
2. Activate a focus mode
3. Deactivate focus mode

Results:
Breaks are paused immediately as focus mode is activated. When it is then deactivated, break resumes after a few seconds (There seems to be this delay for the system to update the flag that is being checked).

### DIFF
--- a/app/utils/dndManager.js
+++ b/app/utils/dndManager.js
@@ -141,14 +141,24 @@ class DndManager extends EventEmitter {
         } catch (e) { wfa = -1 } // getFocusAssist() throw an error if OS isn't windows
         return wfa !== -1 && wfa !== 0
       } else if (process.platform === 'darwin') {
+        const macOSMajorVersion = parseInt(process.getSystemVersion().split('.')[0])
+        let cmd = ''
+        if (macOSMajorVersion >= 26) {
+          cmd = 'defaults read com.apple.controlcenter "NSStatusItem VisibleCC FocusModes"'
+        } else {
+          cmd = 'defaults read com.apple.controlcenter "NSStatusItem Visible FocusModes"'
+        }
         try {
           const asyncExec = this._getOrCreateAsyncExec()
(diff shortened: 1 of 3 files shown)


app/utils/dndManager.js
```diff
@@ -141,14 +141,24 @@ class DndManager extends EventEmitter {
...
         try {
           const asyncExec = this._getOrCreateAsyncExec()
-          const { stdout } = await asyncExec('defaults read com.apple.controlcenter "NSStatusItem Visible FocusModes"')
+          const { stdout } = await asyncExec(cmd)
           if (stdout.replace(/[^0-9a-zA-Z]/g, '') === '1') {
             return true
           }
         } catch (e) {
-          this._logErrorOnce('macos', e)
+          if (!e.message.includes('The domain/default pair of (com.apple.controlcenter, NSStatusItem VisibleCC FocusModes) does not exist')) {
```
hovancik, line 158: Why do we have this `if`?

zhekai-jiang: I commented on the next line. It seems like this entry would just not exist if there is no focus mode enabled. In this case the command returns an error message, but we wouldn't want to log it, I suppose.

hovancik, line 158: Gotcha, I missed `!` and got confused.

hovancik (approved):
Thanks!

PR #1736, opened 2026-02-02
feat: Add custom break message feature

hozaifa1:
Issue: # (no issue was opened for this feature; happy to link one if needed)

### Requirements

- [x]  during development, `node` version specified in [package.json](cci:7://file:///f:/Projects/stretchly/package.json:0:0-0:0) was used (Node 22.20.0 via nvm).
- [x]  package versions and package-lock.json were not changed (`npm install --no-save` was used; lockfile updates only from dependency removal + rebuild).
- [x]  all new code has tests to ensure against regressions (existing suite covers the flows touched).
- [x]  README and CHANGELOG were updated accordingly. *(No docs required because user-facing behavior is self‑explanatory and described in-app.)*
- [x]  after PR is approved, all commits in it are squashed.

### Description of the Change

- Added a `customBreakMessage` setting so users can store a personal message.
- Wired the preferences renderer + HTML to show an input field (with i18n strings) under Settings for editing that message.
- Rendered the saved message in both break and microbreak windows (above the default ideas) when non-empty, sanitized, and hidden otherwise.
- Added English locale strings for the label/help text.
- Adjusted Windows build config: removed the unused `windows-focus-assist` dependency (was breaking builds) and limited `electron-builder` targets to NSIS + portable so Windows users without symlink privileges can build locally.

### Verification Process

1. `npm install` (after removing the failing dependency)  
2. `npm run lint` (passes)  
3. `npm run test` (207/207 passing via Vitest)  
4. `npm start`  
   - Preferences → Settings → entered “Take a deep breath!” → message persisted.  
   - Triggered both mini and long breaks; custom message showed and hid correctly when cleared.  
5. `npm run dist` (succeeds now; generates installer + portable binaries under [dist/](cci:9://file:///f:/Projects/stretchly/dist:0:0-0:0))

### Other information

- Windows builds were previously blocked by symlink + code-signing tool downloads; constraining the targets solved it while still producing installer + portable artifacts.
- Ready to squash on approval if requested.

### DIFF
--- a/app/break-renderer.js
+++ b/app/break-renderer.js
@@ -19,10 +19,19 @@ window.onload = async (event) => {
   document.querySelector('#postpone').onclick = async event =>
     await window.breaks.postponeBreak()
 
+  const customMessage = await window.settings.get('customBreakMessage')
+  const customMessageElement = document.querySelector('.custom-break-message')
+  if (customMessage && customMessage.trim() !== '') {
+    customMessageElement.innerHTML = window.breaks.sanitizeIdea(customMessage)
+    customMessageElement.style.display = 'block'
+  } else {
+    customMessageElement.style.display = 'none'
+  }
+
   document.querySelector('.break-idea').innerHTML = window.breaks.sanitizeIdea(idea[0])
   document.querySelector('.break-text').innerHTML = window.breaks.sanitizeIdea(idea[1])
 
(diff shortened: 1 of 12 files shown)


hovancik:
@hozaifa1 we already allow custom messages via editing the config file, see details here: https://github.com/hovancik/stretchly?tab=readme-ov-file#editing-break-ideas

PR #1766, opened 2026-03-31
Devcontainer improvement

### LINKED ISSUE #1765: [Feature request]: Improving the developer productivity and streamline setup using devcontainers
# Add VS Code Dev Container for zero-friction contributor setup

## Summary

Setting up a local development environment for Stretchly currently requires contributors to manually identify and install a set of system-level dependencies that are not documented anywhere in the repo. On a fresh Debian/Ubuntu machine, running `npm install` fails immediately because `node-desktop-idle-v2` requires `libinput` and `libudev` development headers to compile its native binding, and there is no indication of this anywhere in the README or CONTRIBUTING guide.

A VS Code Dev Container solves this entirely. With a single click ("Reopen in Container"), contributors get a fully working environment where `npm install`, `npm run lint`, `npm test`, and `npm run dev` all work without any manual setup.

## What a Dev Container provides

Every contributor gets the exact same environment regardless of their host OS. The container image pre-installs all Electron runtime libraries, the native build dependencies, and applies the execute bit fix automatically on first creation. SSH agent forwarding, `.gitconfig`, and `.npmrc` are mapped from the host, so authentication and git identity work without any extra steps. The ESLint and Vitest extensions are pre-installed and ESLint is configured to auto-fix on save.

## Impact

The time to go from cloning the repo to a passing `npm test` drops from "figure out which system packages are missing" to approximately two minutes. This is especially valuable for contributors who want to make a small fix or a translation update and would otherwise give up at the setup step.

This is a purely additive change. It introduces no new runtime code and cannot break anything for contributors who do not use it.

navidpadid:
Issue: closes #1765

- [x] issue was opened to discuss proposed changes before starting implementation.
- [x] during development, `node` version specified in `package.json` was used.
- [x] package versions and package-lock.json were not changed.
- [x] app version number was not changed.
- [x] all new code has tests to ensure against regressions.
- [x] README and CHANGELOG were updated accordingly.

### Description of the Change

Adds a VS Code Dev Container so contributors can get a fully working development environment in one click, without manually installing Electron runtime libraries or native build dependencies on their host machine.

**`.devcontainer/Dockerfile`**
- Based on `mcr.microsoft.com/devcontainers/javascript-node:1-22-bookworm`
- Installs all Electron runtime libraries needed for the app and tests (`libasound2`, `libatk-bridge2.0-0`, `libgbm1`, `libgtk-3-0`, `libnss3`, `libxshmfence1`, `xvfb`)
- Includes `libinput-dev` and `libudev-dev` which are required by `node-desktop-idle-v2` to compile its native binding via `node-gyp` — without these, `npm install` fails with `gyp ERR! configure error`

**`.devcontainer/devcontainer.json`**
- Runs `npm install` automatically on container creation, including a `find`-based `chmod u+x` pass on `.bin/` symlink targets to fix a missing execute bit on `cross-env` (shipped without it in the npm tarball), which otherwise causes `npm run lint` to fail with `Permission denied`
- Mounts SSH agent, `~/.ssh`, `~/.gitconfig`, and `~/.npmrc` read-only from the host so Git over SSH and npm auth work without copying credentials into the container
- Forwards port `9222` for remote debugging with `npm run dev`
- Pre-installs the ESLint and Vitest VS Code extensions and configures ESLint to auto-fix on save

**`README.md`**
- Adds a _Dev Container_ section under the contributing guide explaining prerequisites and usage

### Verification Process

Opened the repository in the dev container from scratch and verified:

- `npm install` completes without errors (native modules `node-desktop-idle-v2` and `windows-focus-assist` build successfully)
- `npm run lint` — no offenses
- `npm test` — 315/315 tests pass
- `npm run pack -- --linux` — produces a valid `dist/linux-unpacked/stretchly` binary

### Other information

The `node_modules` directory is kept in a named Docker volume (`stretchly-node-modules`) to avoid re-downloading packages on every container rebuild.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,16 @@
+FROM mcr.microsoft.com/devcontainers/javascript-node:1-22-bookworm
+
+RUN apt-get update \
+  && export DEBIAN_FRONTEND=noninteractive \
+  && apt-get install -y --no-install-recommends \
+    libasound2 \
+    libatk-bridge2.0-0 \
+    libgbm1 \
+    libgtk-3-0 \
+    libinput-dev \
+    libnss3 \
+    libudev-dev \
+    libxshmfence1 \
+    xvfb \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/*
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,47 @@
+{
+  "name": "Stretchly Dev",
+  "build": {
+    "dockerfile": "Dockerfile",
+    "context": ".."
+  },
+  "workspaceFolder": "/workspaces/stretchly",
+  "remoteUser": "node",
+  "updateRemoteUserUID": true,
+  "runArgs": [
+    "--init"
+  ],
+  "remoteEnv": {
+    "SSH_AUTH_SOCK": "/ssh-agent"
+  },
+  "mounts": [
+    "source=stretchly-node-modules,target=/workspaces/stretchly/node_modules,type=volume",
+    "source=${localEnv:SSH_AUTH_SOCK},target=/ssh-agent,type=bind",
+    "source=${localEnv:HOME}/.ssh,target=/home/node/.ssh,type=bind,readonly",
+    "source=${localEnv:HOME}/.gitconfig,target=/home/node/.gitconfig,type=bind,readonly",
+    "source=${localEnv:HOME}/.npmrc,target=/home/node/.npmrc,type=bind,readonly"
+  ],
+  "features": {
+    "ghcr.io/devcontainers/features/git:1": {}
+  },
+  "initializeCommand": "touch ~/.gitconfig ~/.npmrc",
+  "postCreateCommand": "sudo mkdir -p /workspaces/stretchly/node_modules && sudo chown -R node:node /workspaces/stretchly/node_modules && npm install && find node_modules/.bin -maxdepth 1 -type l -exec sh -c 'chmod u+x \"$(readlink -f \"$0\")\"' {} \\;",
+  "forwardPorts": [
+    9222
+  ],
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "dbaeumer.vscode-eslint",
+        "vitest.explorer"
+      ],
+      "settings": {
+        "eslint.validate": [
+          "javascript"
+        ],
+        "editor.codeActionsOnSave": {
+          "source.fixAll.eslint": "explicit"
+        }
+      }
+    }
+  }
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -8,7 +8,8 @@
   "remoteUser": "node",
   "updateRemoteUserUID": true,
   "runArgs": [
-    "--init"
+    "--init",
+    "--network=host"
   ],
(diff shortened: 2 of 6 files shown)


navidpadid:
@hovancik please let me know if there's anything I missed or you have any suggestions. I have tested the setup on Ubuntu and MacOS with Docker. Don't have windows machine to test with WSL, but I believe it should work just fine.

hovancik:
Hi @navidpadid, how does this work? I am not familiar with devcontainer that much. 

The app runs then normally? I mean with this being docker and Stretchly being UI app?

Thanks

navidpadid:
> Hi @navidpadid, how does this work? I am not familiar with devcontainer that much.
> 
> The app runs then normally? I mean with this being docker and Stretchly being UI app?
> 
> Thanks

Hi @hovancik , yes it would be running normally by doing port forwarding. I have added the commands for simplicity to the files. In short, the `npm run dev:devcontainer` would do the trick. I added another command for your testing.

Here is a screenshot of testing command:
Screenshot_20260405_200349


Would you also try to follow the readme and do your tests (UI tests) to confirm if everything works? I have tested this on Linux machine. There could be spots I may have missed.

Thanks.

hovancik:
Hey, I am quite busy, will have to find time to review/try later

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM mcr.microsoft.com/devcontainers/javascript-node:1-22-bookworm
```
hovancik, line 1: could we make it use .nvmrc version? so I don't have to remember update this. Each Electron version is with specific node version so best to keep it in sync.

package.json
```diff
@@ -7,6 +7,7 @@
   "scripts": {
     "start": "electron .",
     "dev": "cross-env NODE_ENV=development electron . --trace-warnings --trace-deprecation --enable-logging --remote-debugging-port=9222",
+    "dev:container": "cross-env NODE_ENV=development electron . --no-sandbox --trace-warnings --trace-deprecation --enable-logging --remote-debugging-port=9223 --remote-allow-origins=*",
```
hovancik, line 10: --remote-allow-origins=* is broad; --remote-allow-origins=http://localhost:9223 would be safer?

hovancik: also, is there reason to use 9223?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,44 @@
...
+    "source=stretchly-node-modules,target=/workspaces/stretchly/node_modules,type=volume",
+    "source=${localEnv:HOME}/.ssh,target=/home/node/.ssh,type=bind,readonly",
+    "source=${localEnv:HOME}/.gitconfig,target=/home/node/.gitconfig,type=bind,readonly",
+    "source=${localEnv:HOME}/.npmrc,target=/home/node/.npmrc,type=bind,readonly"
+  ],
+  "features": {
+    "ghcr.io/devcontainers/features/git:1": {}
+  },
+  "initializeCommand": "touch ~/.gitconfig ~/.npmrc",
+  "postCreateCommand": "sudo mkdir -p /workspaces/stretchly/node_modules && sudo chown -R node:node /workspaces/stretchly/node_modules && npm install && find node_modules/.bin -maxdepth 1 -type l -exec sh -c 'chmod u+x \"$(readlink -f \"$0\")\"' {} \\;",
```
hovancik, line 24: What's this doing?

Issue #355, opened 2019-01-11
Full screen applications disable Stretchly


aidanmischke:
I would prefer not to receive break notifications when running full screen applications as break notifications cause the full screen application to minimize. 

When I play games or run other full screen applications it would be great if Stretchly could set itself to do not disturb or disable itself.

When the full screen application is closed, Stretchly would revert to its previous state and act as per usual.

Ideally this feature could be toggled in the settings section.

Matsku12:
I was coming to write about same thing, but I am suggesting diffrent thing. I used to use app called EyeLeo and it used notifications that came in while gaming, but it did not crash the games like this does. Eyeleo somehow shows notifications over game without it losing the full screen mode. If that cannot be done I would like that there would be some way that the app would keep record how many pauses was skipped and force to keep them all in row before can continue. 

Would be superior if someone would have skills to make this somehow do something with philips hue. Like while gaming there would be no notifications, but philips hue lights would turn to some other color for break and then turn back normal when break is kept.

hovancik:
AFAIK there is no way for electron/nodejs apps to check whether some other app is fullscreen ;/ 

Another problem would be telling when to ignore fullscreen and when not. Ie while gaming it might make sense, but with fullscreen browser probably not. 

I think the easiest way is to simply pause/resume stretchly manually.

Matsku12:
I use these programs mainly for gaming. Eyeleo and this. Both have one "dealbreaker" which makes them unusable for me :/ in eyeleo its afk recognition and with this the problem with breaking full screen.

Would it be possible to add some setting for that? I´d rather would turn this manually off when watching video or such and keep on while gaming. Could you add "ignore fullscreen" setting and problems would be solved?

hovancik:
Unfortunately, the problem is that I know of no way for me to see if other apps are in fullscreen ;/

Matsku12:
What if you make it work allways at the top no matter is or is there not some app in fullscreen?

hovancik:
stretchly is set to be always on top https://github.com/hovancik/stretchly/blob/master/app/main.js#L308 

I guess those other apps manage to get even more on top somehow. I would be lying if I tell I understand how that works

hovancik:
Unfortunately that's not for NodeJS ;/

User486375:
@hovancik Would either of [these](https://stackoverflow.com/questions/39930279/how-can-you-get-information-about-other-apps-running-or-in-focus) Stack Overflow answers be of any use?

[...]

PR #1781, opened 2026-05-17
Add option to pause breaks when another app is in full screen

dkumankov:
Issue: #355

### Requirements

- [x] issue was opened to discuss proposed changes before starting implementation. *(See discussion in #355; design proposal posted as comment before opening this PR.)*
- [x] during development, `node` version specified in `package.json` was used (ie using [nvm](https://github.com/creationix/nvm)).
- [x] package versions and package-lock.json were not changed (`npm install --no-save`).
- [x] app version number was not changed.
- [x] all new code has tests to ensure against regressions.
- [x] README and CHANGELOG were updated accordingly.
- [ ] after PR is approved, all commits in it are [squashed](https://gitbetter.substack.com/p/how-to-squash-git-commits) *(will squash on approval)*

### Description of the Change

Adds an opt-in setting **"Pause breaks while another app is in full screen"** (`monitorFullscreenApp`, default `false`) under Settings, right below the existing "Show breaks even in Do Not Disturb mode" toggle.

When enabled, break scheduling is paused while another application is presenting in fullscreen, and resumes when fullscreen ends — symmetrical to how `monitorDnd` works today.

**Architecture.** A new `FullscreenAppManager` (`app/utils/fullscreenAppManager.js`) mirrors `DndManager`:
- extends `EventEmitter`
- interval polling (2s)
- `start()` / `stop()` lifecycle
- exposes `isOnFullscreenApp` boolean
- emits `fullscreenAppStarted` / `fullscreenAppFinished`

`BreaksPlanner` constructs it, subscribes to its events, and exposes a `fullscreenApp(shouldUse)` method symmetrical to `doNotDisturb(shouldUse)`. `main.js` adds `isOnFullscreenApp` to the existing guard expressions (`onSuspendOrLock`, `trayIconPath`, the two tray-menu branches, `resumeBreaks`) so the manager participates in all the same "is Stretchly currently suppressed" checks as DND and app exclusions.

**Cross-platform detection — no new dependencies:**

| Platform | Mechanism |
|---|---|
| Windows | `getFocusAssist()` from the already-present `windows-focus-assist` (state 3 = D3D fullscreen, 4 = presentation mode) **OR** a persistent PowerShell worker that compares the foreground window's `GetWindowRect` to its monitor's `rcMonitor` — catches borderless-windowed games and browser F11 that the notification-state API alone reports as `BUSY`. |
| macOS | `osascript` reading `AXFullScreen` of the frontmost window — same shell-out pattern as `dndManager`'s macOS path. |
| Linux (X11) | `xdotool getactivewindow` + `xprop _NET_WM_STATE` looking for `_NET_WM_STATE_FULLSCREEN`. |
| Linux (Wayland) | Graceful no-op (logged once at startup); no portable client API for inspecting other windows' fullscreen state. |
| Other | Logs an "unsupported" notice once and returns `false`. |

i18n strings added across all 49 locales that currently expose the `preferences.settings` namespace (5 incomplete locales — `arz`, `bn`, `cy`, `eu`, `sr` — fall back to English at runtime, matching Stretchly's existing pattern).

### Verification Process

**Automated**
- `npm run lint` — clean.
- `npm run test` — passes, including 8 new unit tests in `test/fullscreenAppManager.js` covering: default-off state, `start()`/`stop()` lifecycle, `start()` idempotency, rising-edge / falling-edge emission, `_detectFullscreen` no-op when disabled, log-once behavior on detection errors.
- Locale JSON validation (`test/translations.js`) — all 54 files valid.

**Manual (Windows 11)**

| Action | Expected | Observed |
|---|---|---|
| Enable `monitorFullscreenApp` in Preferences | log shows `starting fullscreen app monitoring` | ✅ |
| Open a borderless-fullscreen app (Crimson Desert) | log shows `pausing breaks because a fullscreen app is active` within ~2 s; tray icon switches to paused style | ✅ |
| Exit the fullscreen app | log shows `resuming breaks after fullscreen app is no longer active`; tray icon returns to normal | ✅ |
| Start a Stretchly break while the setting is on, with a fullscreen app already open | break still shows; fullscreen-app handler bails because `scheduler.reference === 'finishBreak'` | ✅ |
| Toggle the setting off while a fullscreen app is active | manager stops polling; planner resumes (mirrors `doNotDisturb(false)`) | ✅ |

**Regression checks:** the existing DND / app-exclusion / suspend-lock paths continue to behave the same way — all guards that previously read `isPaused || isOnDnd || isSchedulerCleared` now additionally include `isOnFullscreenApp`, but the OR-of-booleans is purely additive.

### Other information

- The persistent PowerShell worker is only spawned on Windows when the setting is enabled; it's terminated by `stop()`. No process is created at all on macOS/Linux.
- The two-signal detection on Windows (`SHQueryUserNotificationState` OR bounds-equals-monitor) was added after observing that modern games like Crimson Desert run in borderless-windowed mode rather than exclusive D3D fullscreen, so the notification-state API alone misses them.
- Open to switching to the `"0-fullscreen"`-token-inside-`appExclusions` design you sketched in [issue #355's 2023 comment](https://github.com/hovancik/stretchly/issues/355#issuecomment-1474842867) if you prefer that integration; the manager itself stays the same, only the wiring changes.


app/utils/fullscreenAppManager.js
```diff
@@ -0,0 +1,268 @@
...
+import log from 'electron-log/main.js'
+import { getFocusAssist } from 'windows-focus-assist'
+import { exec, spawn } from 'node:child_process'
+import { promisify } from 'node:util'
+
+// SHQueryUserNotificationState values surfaced by windows-focus-assist that
+// indicate another app is presenting in full screen.
+//   3 = QUNS_RUNNING_D3D_FULL_SCREEN (exclusive-fullscreen D3D app)
+//   4 = QUNS_PRESENTATION_MODE       (PowerPoint slideshow, fullscreen video)
+const WINDOWS_FULLSCREEN_STATES = new Set([3, 4])
```
hovancik: How does this work? I am seeing in https://github.com/bitdisaster/windows-focus-assist/blob/master/lib/index.js and they have -2..2

dkumankov: Thanks @hovancik - I see what you're getting at, and it actually nudged me toward a much cleaner way out of the whole PowerShell mess that was already bothering me.

[`windows-notification-state`](https://github.com/felixrieseberg/windows-notification-state) is by the same Felix Rieseberg as `windows-focus-assist` (already a dep here for DND) and gives us exactly this in one synchronous call. Subprocess, inline C# and stdout parser all gone - FullscreenAppManager finally reads like a clean little file.

hovancik: Hey, thanks for response.  Is this human or an AI assistant? I ask because this reply sounds very AI and I find async talk with AI over PR not effective..No offense meant

dkumankov: image

app/locales/ajp.json
```diff
@@ -103,6 +103,7 @@
       "fullscreen": "شاشة عرض كاملة",
       "monitorIdleTime": "مراقبة وقت خمول النظام (الإستراحات تتوقف خلال وقت الخمول).",
       "monitorDnd": "إظهار الإستراحات حتى خلال وضع عدم الإزعاج",
+      "monitorFullscreenApp": "إيقاف الاستراحات مؤقتاً عندما يكون تطبيق آخر بملء الشاشة",
```
hovancik: can you please remove any translation that is not your native language or English? 

I prefer natives do them over AI/machine translators

hovancik:
so how should this behave if user sets "Show breaks even in Do not Disturb mode" and "pause in fullscreen"? 

in my experience, some apps set "do not disturb" when they go "fullscreen".

app/utils/fullscreenAppManager.js
```diff
@@ -0,0 +1,166 @@
...
+      // is no reliable, portable detection — fall back to "not fullscreen".
+      if (!this._waylandWarned) {
+        log.info('Stretchly: fullscreen app detection is not supported on Wayland')
+        this._waylandWarned = true
+      }
+      return false
+    }
+    try {
+      const asyncExec = this._getOrCreateAsyncExec()
+      const { stdout: idOut } = await asyncExec('xdotool getactivewindow')
```
hovancik, line 128: `xdotool` and `xprop` - are those on every linux system?

app/locales/ajp.json
dkumankov: Yes, sure

dkumankov:
> so how should this behave if user sets "Show breaks even in Do not Disturb mode" and "pause in fullscreen"?
> 
> in my experience, some apps set "do not disturb" when they go "fullscreen".

They're two independent checks - show breaks even in DnD just ignores the DnD state, while the fullscreen option looks at the actual window. So with both on you still get breaks during plain DnD, but not while an app is genuinely fullscreen.

app/utils/fullscreenAppManager.js
dkumankov: Yes, my bad - I assumed x11-utils was always there (it was in my Ubuntu VM). There's a try/catch so missing packages won't crash anything, but honestly there's not much point if they're not installed by default and few people will bother installing extra binaries just for this. Maybe we hide the checkbox on Linux and i cut the related code?

hovancik:
> > so how should this behave if user sets "Show breaks even in Do not Disturb mode" and "pause in fullscreen"?
> > in my experience, some apps set "do not disturb" when they go "fullscreen".
> 
> They're two independent checks - show breaks even in DnD just ignores the DnD state, while the fullscreen option looks at the actual window. So with both on you still get breaks during plain DnD, but not while an app is genuinely fullscreen.

TBH I am still not sure about this. Need to think more. 

Also, what OS did you test on? I am trying now on WIndows 11 and making Firefox fullscreen does not pauses Stretchly. Also seeing crashes when clicking on checkbox in Preferences.

PR #1804, opened 2026-06-18
Disable text selection on break screens

### LINKED ISSUE #1805: [Bug]: Text selection on break screens allows escaping the break (macOS force-click dictionary)
### Version

I'm using the current `trunk` (affects 1.21.0).

### Known issues / Existing issues / Advanced Preferences

I've checked Known issues, Existing issues and Advanced Preferences — no match.

### Operating system

macOS

### Operating System Version

macOS 26 (Darwin 25.5.0)

### Reproduction steps

1. Start a break that shows the fullscreen break overlay.
2. On the break screen, double-click or drag to select one of the words of the break text.
3. Force-click (firm press on the trackpad) the selected word to bring up the macOS dictionary/Look Up popover, then click **Open in Dictionary**.
4. The Dictionary app opens and you are returned to the desktop — the break overlay has been bypassed without skipping/finishing the break.

### Expected Behavior

Break-screen content should not be selectable or otherwise interactive. There should be no way to leave the break overlay other than the intended skip/finish controls (and strict mode should block even those).

### Actual Behavior

Break-screen text is selectable. On macOS, selecting a word exposes the force-click dictionary popover, and "Open in Dictionary" launches another app, dropping the user back to the desktop and escaping the break. This is especially relevant to strict-mode breaks, whose whole purpose is to be inescapable.

### Additional information

Fix proposed in an attached PR: make all break-screen content non-selectable in `break.css` (`user-select: none`), which removes the selection the force-click dictionary path depends on.

### Code of Conduct

I agree to follow this project's Code of Conduct.

githappens:
Closes #1805

### Requirements

- [ ]  issue was opened to discuss proposed changes before starting implementation. _Issue #1805 was opened alongside this PR rather than strictly before implementation — happy to discuss before it is merged._
- [ ]  during development, `node` version specified in `package.json` was used. _This is a CSS-only change with no build/runtime Node dependency._
- [ ]  all new code has tests to ensure against regressions. _The change is a `user-select: none` style rule; selectability/force-click behavior is not unit-testable without a macOS UI integration harness, which the project does not currently have._
- [x] `npm run lint` reports no offenses. _StandardJS lints JS only; no JS changed and the full lint run is clean._
- [ ] `npm run test` is error-free. _All 250 tests pass; a few test files fail to collect locally due to missing optional packages in my environment, unrelated to this change._
- [ ]  after PR is approved, all commits in it are squashed. _Will squash on approval._

### Description of the Change

On the break screen, the text content is selectable. On macOS this is an escape hatch out of the break overlay: select a word → force-click (firm trackpad press) to bring up the dictionary / Look Up popover → **Open in Dictionary** → the Dictionary app opens and you are dropped back to the desktop, bypassing the break without skipping or finishing it. This is especially relevant to strict-mode breaks, whose purpose is to be inescapable.

The fix makes all break-screen content non-selectable in `app/css/break.css` (loaded by both `break.html` and `microbreak.html`, so it covers Mini and Long breaks):

- `* { -webkit-user-select: none; user-select: none; cursor: default; }` removes the selection that the force-click dictionary path depends on.
- `::selection { background: transparent }` as a fallback.

The postpone/skip/finish actions are `<a>` elements without an `href`, so they already used the default arrow cursor before this change; the universal `cursor: default` keeps that behavior unchanged.

CSS-only, scoped to the break windows. No behavior change beyond text no longer being selectable during breaks.

### Verification Process

- macOS: triggered Mini and Long breaks and confirmed break text can no longer be selected (double-click, drag) and the force-click dictionary → "Open in Dictionary" escape no longer works.
- Confirmed the postpone/skip/finish actions keep their original cursor and work as before.
- `npm run lint` clean.

### Other information

The macOS force-click dictionary was the specific reported vector. Disabling selection is cross-platform and harmless elsewhere. Right-click context menu and image drag-out are adjacent hardening opportunities but are intentionally out of scope for this change.


hovancik (changes requested):
LGTM, with one issue: cursor on break actions (postpone etc) is wrong. You can compare with current behavior.

hovancik (approved):

Issue #1628, opened 2025-08-14
[Feature request]: Get current state from CLI


dkaszews:
### Version

- [x] I'm using version 1.17.2

### Existing requests

- [x] I've checked Existing requests

### Advanced Preferences

- [x] I've checked Advanced Preferences

### What operating system are you using?

Linux

### Operating System Version

Linux Cuddles-Bobba 6.15.9-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 02 Aug 2025 01:20:06 +0000 x86_64 GNU/Linux

### Problem description

I would like to add Stretchly to my status bar (polybar to be exact). While it can be added as a tray icon, I want immediate access to more information like current pause state, time to next mini/long break, etc. Polybar (and other text statusbars) work by periodically running scripts and displaying their outputs, but `stretchly` CLI command does not seem to have a status command.

### Possible solution

Add `stretchly status` command which displays relevant dynamic state information in script-friendly format, e.g. JSON. Example output:

```json
{
    "seconds_paused": 0,  // implies unpaused
    "seconds_to_mini_break": 120,
    "seconds_to_long_break": 3600
}
```

### Alternative solutions

Some of this information can already be deduced from logs and config, e.g.:
```
[2025-08-14 22:18:24.147] [info]  Stretchly: finishing Mini Break (shouldPlanNext: true)
```

combined with
```json
	"microbreakInterval": 300000,
```

And current date `2025-08-14 22:21:45.481`, you can calculate the next mini break to be in:

```
2025-08-14 22:21:45.481 - 2025-08-14 22:18:24.147 = 3min21s334ms
300000ms - 3m21s334ms = 5min - 3m21s334ms = 1m38s666ms
```

But the logs show not be considered a stable API, plus may fall apart on some edge cases (e.g. mini break timer resetting due to long break on pauses).

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

hovancik:
Thanks for the feedback, should be doable :)

jeregon1:
I'd be willing to work on it, even though it would be my first time contributing to Stretchly, it seems an easy task. I've seen the command line options are in [main.js](https://github.com/hovancik/stretchly/blob/trunk/app/main.js)

dkaszews:
@jeregon1 I've only done cursory research to implement `stretchly --hello-world`:

* Add command to `allCommands` in `app/utils/commands.js`
[...]

PR #1811, opened 2026-06-23
Feature/cli status

jeregon1:
Issue: #1628

### Requirements

* [x] issue was opened to discuss proposed changes before starting implementation.
* [x] during development, `node` version specified in `package.json` was used (v22+).
* [x] package versions and package-lock.json were not changed.
* [x] app version number was not changed.
* [x] all new code has tests to ensure against regressions.
* [x] `npm run lint` reports no offenses.
* [x] `npm run test` is error-free (24 new tests pass).
* [x] README and CHANGELOG were updated accordingly.
* [x] after PR is approved, all commits in it are squashed.

### Description of the Change

Adds a `stretchly status` CLI command that shows the current break state without opening the GUI. Supports both human-readable text output and `--json` / `-j` for machine parsing (integration with polybar, i3status, etc.).

**States detected:**
- `no_active_break` — shows time remaining until the next mini and long break
- `active_break` — shows break type (mini/long) and time until it ends
- `paused` — shows that breaks are paused
- Unavailable — when no Stretchly instance is running

**Snapshot-based architecture:** The running instance periodically writes a JSON snapshot file. Any second instance reads this file to display status, avoiding IPC complexity and keeping the first instance unmodified.

### Verification Process

- `npm run lint` — clean
- `npm run test` — 24 new tests + 32 existing = 56 passed
- `npm run pack` — builds successfully on Windows
- Tested edge cases: paused, active mini break, active long break, no active break, missing planner, missing settings
- AI-assisted contribution


hovancik:
Hi and thanks for the PR :) 

Can you explain what complexities do you mean here? 

> Snapshot-based architecture: The running instance periodically writes a JSON snapshot file. Any second instance reads this file to display status, avoiding IPC complexity and keeping the first instance unmodified.

jeregon1:
Thank you for reviewing,

It's my first contribution to the project, so probably I don't understand it well enough so that I'm right on what I'm going to say. 
But, thinking of a way to implement the status subcommand with AI I didn't know how to make the first instance answer to the second one (that's the IPC complexity I was referring). 
So the option of the status snapshot looked simple and effective enough.

I'll really appreciate any feedback <3

hovancik:
Ah, gotcha, you mean sending message from 1st to 2nd instance. That's right, we don't have that yet, but writing file every 10 seconds seems excessive. 

I think we should use `statusMessages.js` to generate the cli status, there's a lot of logic already over there. 

We need to decide: how are we going to map it? Do we want to use localized message or english? 

What do you think?

jeregon1:
I believe this command's main purpose is to be used with the  --json flag, i.e. people using this will be mainly for scripting purposes, as the person who created the #1628 stated. 

In that case, I think it's not necessary to add localization.

And yes, I didn't notice the `statusMessages.js` file, maybe we could put all the logic for this subcommand there.

hovancik:
So plan should be: 
- change `statusMessages.js` to also generate cli status message (not sure whether some abstraction or duplication would be better)
- save to file when second instance asks for it and second instance reads it

Then we can nitpick about specific wording in messages

